PR #1, opened 2019-11-27
Feature/osd

szenzaro:
Adds first OpenSeadragon component supporting manifests


src/app/app.config.ts
```diff
@@ -64,6 +64,7 @@ export interface UiConfig {
 
 export interface EditionConfig {
     title: string;
+    manifestURL: string; // TODO: check name and position for this property
```
ChiaraDipi: I would move `manifestUrl` configuration from `EditionConfig` interface to `FileConfig`. 
Also, why not use camel case for property name?

src/assets/config/edition_config.json
```diff
@@ -1,3 +1,4 @@
 {
-	"title": "My Digital Edition"
+	"title": "My Digital Edition",
+	"manifestURL": "https://www.e-codices.unifr.ch/metadata/iiif/ubb-B-VI-0016/manifest.json"
```
ChiaraDipi: Same as previous comment: I would put this config among files configuration.

src/app/app.config.ts
szenzaro: Ok for the `FileConfig`. 

Usually it is considered camel case since URL is an acronym

PR #9, opened 2020-01-06
Feature/parse xml attributes

giuliac89:


src/app/models/evt-models.ts
```diff
@@ -4,6 +4,10 @@ import { GenericElementData } from './parsed-elements';
 export type XMLElement = HTMLElement;
 export type OriginalEncodingNodeType = XMLElement;
 
+export interface AttributesData {
+    [key: string]: string ;
```
ChiaraDipi: Remove space before ";"

szenzaro: Use automatic formatter

src/app/utils/xml-utils.ts
```diff
@@ -20,3 +20,12 @@ export function parseXml(xmlStr: string): XMLElement {
   }
   throw new Error('No XML parser found');
 }
+
+export function parseAttributes(xml: XMLElement): AttributesData {
+  const attributes: AttributesData = {};
+  Array.from(xml.attributes).forEach((attr) => {
```
szenzaro: Replace `(attr)` with `({name, value})` and use them directly

ChiaraDipi: You can also remap attributes with `xml` namespace, such as `xml:id` or `xml:lang`, in order to remove the `xml:` prefix. 
Moreover, use this function in `GenericParserService` instead of its private `getAttributes`: they have the same purpose.

PR #10, opened 2020-01-06
Feature/text parser optimization

giuliac89:


src/app/services/xml-parsers/generic-parser.service.ts
```diff
@@ -4,6 +4,7 @@ import { map, scan } from 'rxjs/operators';
 import { NoteComponent } from 'src/app/components/note/note.component';
 import { XMLElement } from 'src/app/models/evt-models';
 import { isNestedInElem, xpath } from 'src/app/utils/dom-utils';
+import { normalizeContentSpaces } from 'src/app/utils/xml-utils';
```
ChiaraDipi: Use relative path for imports (I know, there are other wrong imports, we will fix them in the future!)

src/app/utils/xml-utils.ts
```diff
@@ -20,3 +20,7 @@ export function parseXml(xmlStr: string): XMLElement {
   }
   throw new Error('No XML parser found');
 }
+
+export function normalizeContentSpaces(textContent: string) {
```
szenzaro: What about calling this `replaceMultispaces`

PR #12, opened 2020-01-08
Feature/named entities lists extraction

ChiaraDipi:


src/app/models/evt-models.ts
```diff
@@ -15,3 +15,50 @@ export interface PageData {
...
+export type AttributesData = Array<{ key: string; value: string }>;
+
+export interface OriginalEncodingData {
+    originalEncoding: OriginalEncodingNodeType;
+}
+
+export type NamedEntityType = 'person' | 'place' | 'org' | 'relation' | 'event' | 'generic';
+export interface NamedEntitiesList extends GenericElementData {
+    id: string;
+    label: GenericElementData | string;
```
szenzaro: Weird that a label isn't always a string. Check it.

src/app/models/evt-models.ts
```diff
@@ -15,3 +15,50 @@ export interface PageData {
...
+export interface NamedEntity extends GenericElementData {
+    id: string;
+    label: NamedEntityLabel;
+    namedEntityType: NamedEntityType | 'personGrp';
+    content: NamedEntityInfo[];
+    occurrences: string[]; // TODO: evaluate which type assign
+    originalEncoding: OriginalEncodingNodeType;
+}
+
+export type NamedEntityLabel = GenericElementData | string | Array<GenericElementData | string>;
```
szenzaro: A label should be a string. When are other types needed?

src/app/services/xml-parsers/named-entities-parser.service.ts
```diff
@@ -0,0 +1,343 @@
...
+  );
+
+  public readonly namedEntities$ = combineLatest([
+    this.persons$,
+    this.places$,
+    this.organizations$,
+    this.relations$,
+    this.events$,
+  ]).pipe(
+    map(ne => ({
```
szenzaro: `ne` can be `[persons, places, organizations, relations, events]`

src/app/services/xml-parsers/named-entities-parser.service.ts
```diff
@@ -0,0 +1,343 @@
...
+    private genericParserService: GenericParserService,
+  ) {
+  }
+
+  private parseLists(document: XMLElement) {
+    const listsToParse = this.getListsToParseTagNames();
+    const parsedLists: NamedEntitiesList[] = [];
+    let parsedEntities: NamedEntity[] = [];
+    let parsedRelations: Relation[] = [];
+    if (listsToParse) {
```
szenzaro: no need of this check, `getListsToParseTagNames` always returns an array (maybe empty but not undefined or null)

src/app/services/xml-parsers/named-entities-parser.service.ts
```diff
@@ -0,0 +1,343 @@
...
+      const lists = document.querySelectorAll<XMLElement>(listsToParse.toString());
+      lists.forEach((list) => {
+        // We consider only first level lists; inset lists will be considered differently
+        if (!isNestedInElem(list, list.tagName)) {
+          const parsedList = this.parseList(list);
+          parsedList.id = parsedList.id;
```
szenzaro: ???

src/app/services/xml-parsers/named-entities-parser.service.ts
```diff
@@ -0,0 +1,343 @@
...
+    events: 'listEvent',
+  };
+
+  constructor(
+    private editionDataService: EditionDataService,
+    private genericParserService: GenericParserService,
+  ) {
+  }
+
+  private parseLists(document: XMLElement) {
```
szenzaro: I suggest to refactor this as
```
  private parseLists(document: XMLElement) {
    const listsToParse = this.getListsToParseTagNames();
    // We consider only first level lists; inset lists will be considered
    const lists = Array.from(document.querySelectorAll<XMLElement>(listsToParse.toString()))
      .filter((list) => !isNestedInElem(list, list.tagName))
      .map((l) => this.parseList(l));

    return {
      lists,
      entities: lists.map(({ content }) => content).reduce((a, b) => a.concat(b), []),
      relations: lists.map(({ relations }) => relations).reduce((a, b) => a.concat(b), []),
    };
  }
```

src/app/services/xml-parsers/named-entities-parser.service.ts
```diff
@@ -0,0 +1,343 @@
...
+      case 'personGrp':
+        return this.parsePersonGroup(xml);
+      case 'place':
+        return this.parsePlace(xml);
+      case 'org':
+        return this.parseOrganization(xml);
+      case 'event':
+        return this.parseEvent(xml);
+      default:
+        console.log('Warning! Parser not implemented for this element in list', xml);
```
szenzaro: No `console.log`, consider `console.error` instead until better error handling

src/app/services/xml-parsers/named-entities-parser.service.ts
```diff
@@ -0,0 +1,343 @@
...
+
+  private parseGenericEntity(xml: XMLElement): NamedEntity {
+    const elId = xml.getAttribute('xml:id') || xpath(xml);
+    const entity: NamedEntity = {
+      type: GenericElementComponent, // TODO: Set NamedEntitiesComponent
+      id: elId,
+      originalEncoding: xml,
+      label: this.genericParserService.parse(xml),
+      namedEntityType: this.getEntityType(xml.tagName),
+      content: [],
```
szenzaro: `content: Array.from(xml.childNodes).map((subchild: XMLElement) => this.parseEntityInfo(subchild))`

and remove line 178

src/app/services/xml-parsers/named-entities-parser.service.ts
```diff
@@ -0,0 +1,343 @@
...
+      attributes: this.parseAttributes(xml),
+      occurrences: [],
+    };
+    xml.childNodes.forEach((subchild: XMLElement) => entity.content.push(this.parseEntityInfo(subchild)));
```
szenzaro: Remove line following previous comment

src/app/services/xml-parsers/named-entities-parser.service.ts
ChiaraDipi: Oops, it's something I forgot to delete: before it was written differently and made more sense.

src/app/services/xml-parsers/named-entities-parser.service.ts
ChiaraDipi: Do you mean that I should define a more specific type for `ne`?

src/app/models/evt-models.ts
ChiaraDipi: I have probably over-complicated the situation. I was try to handle all the different type of xml encoding we can face when using TEI encoded files. Details of an entity, such as `<person>` can be encoded in many different ways, both structured and unstructured; it can even be defined as an empty element. E.g.:
```
<person xml:id="zh-tw_LB"/>

<person xml:id="pp1">
   Person pp1
</person>

<person xml:id="cwn">
   <p>Customer WN</p>
</person> 

<person xml:id="Acaritus">
   <persName>
      <forename>Acaritus</forename>
   </persName>
   <sex>M</sex>
</person>

<person xml:id="AAAAATEST">
   <persName>
      <forename>Edward</forename>
      <forename>George</forename>
      <surname type="linked">Bulwer-Lytton</surname>,
      <roleName>Baron Lytton of
         <placeName>Knebworth</placeName>
      </roleName>
   </persName>
   <sex>M</sex>
   <occupation>syndicus et procurator communium et universitatum de Marola et de
      Matrono</occupation>
</person>
```
But you are right: the label is the text used to identify the entity, thus it should probably be a simple string, and I can handle all the possible cases differently, by just extracting the `textContent`. 

This make me think that property of entities (such as information encoded in attributes and child nodes) should be more defined, maybe by just using possible child nodes from `namesdates` module for the moment.

src/app/models/evt-models.ts
ChiaraDipi: If the list element has the child node `<head>` I parse this and set as the label. Otherwise I use the `@type` attribute (that is a `string`). This is because I wanted to save possible formatting within `<head>` element. However, I've checked again the TEI guidelines and all the example of `<head>` within a list only contain text. Thus, we can probably switch to a more simple `string` type for label!

src/app/services/xml-parsers/named-entities-parser.service.ts
szenzaro: no, you can rewrite the line from
```
map(ne => ...)
```
to
```
map(([persons, places, organizations, relations, events]) => ...)
```

PR #15, opened 2020-01-13
Feature/link viewer to image view

szenzaro:
Show OpenSeadragon on image view mode


src/app/panels/image-panel/image-panel.component.ts
```diff
@@ -1,4 +1,5 @@
 import { Component } from '@angular/core';
+import { AppConfig } from 'src/app/app.config';
```
ChiaraDipi: Use relative paths

PR #19, opened 2020-01-19
Feature/witnesses list extraction

giuliac89:


src/app/services/xml-parsers/witnesses-parser.service.ts
```diff
@@ -0,0 +1,123 @@
...
+  }
+
+  private parseWitnesses(list: XMLElement, parsedList: WitnessesList) {
+    Array.from(list.querySelectorAll<XMLElement>(this.tagNamesMap.witTagName))
+      .map((wit) => {
+        const parsedWit = this.parseWitness(wit);
+        parsedList.witnesses[parsedWit.id] = parsedWit;
+        parsedList.indexes.witnesses.push(wit.getAttribute('xml:id'));
+        if (parsedWit.group === undefined) {
+          parsedList.indexes.encodingStructure.push(wit.getAttribute('xml:id'));
```
szenzaro: Avoid side effects Inside maps

src/app/services/xml-parsers/witnesses-parser.service.ts
```diff
@@ -0,0 +1,123 @@
...
+      attributes: this.genericParserService.parseAttributes(wit),
+      content: this.parseWitnessContent(wit),
+      group: this.parseParentGroupId(wit),
+    };
+  }
+
+  private parseWitnessContent(wit: XMLElement): Description {
+    const content = [];
+    Array.from(wit.childNodes)
+      .map((child: XMLElement) => content.push(this.genericParserService.parse(child)));
```
szenzaro: Just use map, non need of side effects

src/app/services/xml-parsers/witnesses-parser.service.ts
```diff
@@ -0,0 +1,123 @@
...
+    const groupEl = Array.from(list.children)
+      .find((child) => child.nodeName === groupTagName);
+
+    return groupEl ? groupEl.textContent : undefined;
+  }
+
+  private parseGroupContent(list: XMLElement): Description {
+    const content = [];
+    Array.from(list.children)
+      .filter(() => Object.keys(this.tagNamesMap))
```
szenzaro: Your intentions are unclear

src/app/models/evt-models.ts
```diff
@@ -84,3 +84,24 @@ export type Relation = OriginalEncodingData & {
 };
 
 export type Description = GenericElementData[];
+export interface WitnessesList {
+    witnesses: Map<Witness>;
```
szenzaro: Could you elaborate on the choice for this structure?

giuliac89: Each witness inside `witnesses` is a key-value pair. In this way I can quickly retrieve the data I need, but probably in this case it's better to use `[key: string]: string;`? I can't figure out exactly what is best to use in this case.

src/app/services/xml-parsers/witnesses-parser.service.ts
giuliac89: I apply a filter that retrieve only the elements in the `tagNamesMap` (`listWit` and `witness`) to get the content of a specific group, that is which witnesses or groups belong to the current group. How can I make this code clearer?

src/app/services/xml-parsers/structure-xml-parser.service.ts
```diff
@@ -73,6 +75,8 @@ export class StructureXmlParserService {
       console.log(pages);
     }
 
+    this.witParserService.parsedList$.subscribe(); // TODO: TEMP to test parser
```
szenzaro: Still TEMP?

src/app/services/xml-parsers/witnesses-parser.service.ts
szenzaro: You are actually filtering by empty vs non empty `this.tagNamesMap` keys. 
If that is what you want, make it explicit `Object.keys(this.tagNamesMap).lenght > 0`.

src/app/services/xml-parsers/witnesses-parser.service.ts
```diff
@@ -0,0 +1,120 @@
...
+
+  private parseGroupName(list: XMLElement) {
+    const groupTagName = 'head';
+    const groupEl = Array.from(list.children)
+      .find((child) => child.nodeName === groupTagName);
+
+    return groupEl ? groupEl.textContent : undefined;
+  }
+
+  private parseGroupContent(list: XMLElement): Description {
```
szenzaro: Rewrite as 
```  private parseGroupContent(list: XMLElement): Description {
   return  Array.from(list.children)
      .filter(() => Object.keys(this.tagNamesMap))
      .map((child) => child.getAttribute('xml:id'));
  }```

File: src/app/services/xml-parsers/witnesses-parser.service.ts [hunk 7]
```diff
@@ -0,0 +1,120 @@
...
+      .map((child) => content.push(child.getAttribute('xml:id')));
+
+    return content;
+  }
+
+  private parseParentGroupId(element: XMLElement) {
+    const parentEl = element.parentElement;
+
+    if (parentEl.tagName !== this.tagNamesMap.witListTagName) {
+      this.parseParentGroupId(parentEl);
```
szenzaro: missing `return`?
You can also write this in for-loop form to avoid recursion

src/app/services/xml-parsers/witnesses-parser.service.ts
```diff
@@ -0,0 +1,120 @@
...
+    }
+
+    if (isNestedInElem(parentEl, parentEl.tagName)) {
```
szenzaro: You just found the parent walking up the parent-child structure, shouldn't it be always true?

src/app/services/xml-parsers/witnesses-parser.service.ts
```diff
@@ -0,0 +1,120 @@
...
+      return parentEl.getAttribute('xml:id') || xpath(parentEl);
+    }
+  }
```
szenzaro: Missing base cases. What happen when it is not nested and `parentEl.tagName === this.tagNamesMap.witListTagName`?

src/app/services/xml-parsers/witnesses-parser.service.ts
```diff
@@ -0,0 +1,120 @@
...
+      group: this.parseParentGroupId(list) || undefined,
+    };
+  }
+
+  private parseGroupName(list: XMLElement) {
+    const groupTagName = 'head';
+    const groupEl = Array.from(list.children)
+      .find((child) => child.nodeName === groupTagName);
+
+    return groupEl ? groupEl.textContent : undefined;
```
szenzaro: you can write `return groupEl && groupEl.textContend`

src/app/services/xml-parsers/witnesses-parser.service.ts
```diff
@@ -0,0 +1,120 @@
...
+    parsedList.indexes.groups.push(parsedGroup.id);
+  }
+
+  private parseWitnessGroup(list: XMLElement): WitnessGroup {
+    return {
+      id: list.getAttribute('xml:id') || xpath(list),
+      name: this.parseGroupName(list) || replaceNotWordChar(list.getAttribute('xml:id')) || xpath(list),
+      attributes: this.genericParserService.parseAttributes(list),
+      content: this.parseGroupContent(list),
+      group: this.parseParentGroupId(list) || undefined,
```
szenzaro: `parseParentGroupId(list) || undefined` is equal to `parseParentGroupId(list)`

src/app/utils/xml-utils.ts
```diff
@@ -28,3 +28,7 @@ export function replaceMultispaces(textContent: string) {
 export function replaceNewLines(textContent: string) {
   return replaceMultispaces(textContent.trim().replace(/\n/g, ' '));
 }
+
+export function replaceNotWordChar(textContent: string) {
+  return textContent ? textContent.replace(/[\W_]/, ' ') : undefined;
```
szenzaro: `return textContent || textContent.replace(/[\W_]/, ' ')`

src/app/models/evt-models.ts
```diff
@@ -84,3 +84,24 @@ export type Relation = OriginalEncodingData & {
...
+    originalGroup: Map<WitnessGroup>;
+    customGroup?: [];
+    indexes: Map<string[]>;
```
szenzaro: Probably redundant (`indexes` =? `Object.keys(witnesses))

src/app/utils/xml-utils.ts
giuliac89: It should be `return textContent && textContent.replace(/[\W_]/, ' ')`, right?

src/app/services/xml-parsers/witnesses-parser.service.ts
giuliac89: I need this condition because in an encoding like below, the first `<listWit>` element, which includes all the others, should not be considered a group.
```
<listWit>
  <listWit xml:id="rec1_wits"></listWit>
  <listWit xml:id="rec2_wits"></listWit>
  <listWit xml:id="rec3_wits"></listWit>
<listWit>
```

src/app/services/xml-parsers/witnesses-parser.service.ts
giuliac89: This case will never be considered, because I never consider the first `<listWit>` (see the previous comment)

src/app/services/xml-parsers/witnesses-parser.service.ts
giuliac89: Do you think it's better to avoid recursion in this case?

szenzaro: Recursion is not a problem, I would go for a function with clear return paths

src/app/services/xml-parsers/witnesses-parser.service.ts
szenzaro: Maybe comment the code about this choice

src/app/models/evt-models.ts
szenzaro: go for the Map, other choices will be considered later

src/app/models/evt-models.ts
```diff
@@ -83,3 +83,23 @@ export type Relation = OriginalEncodingData & {
 };
 
 export type Description = GenericElementData[];
+export interface WitnessesList {
+    witnesses: Map<Witness>;
+    originalGroup: Map<WitnessGroup>;
+    customGroup?: [];
```
szenzaro: check type

PR #20, opened 2020-01-20
Switch to components registry for dynamic component

ChiaraDipi:
Change parsing output to use component registry in content viewer component, in order to avoid too many dependencies in parsing services.


src/app/components/content-viewer/content-viewer.component.ts
```diff
@@ -22,8 +22,16 @@ export class ContentViewerComponent implements OnDestroy {
...
 
+  constructor(
+    private componentRegister: ComponentRegisterService,
+  ) {
+  }
   // tslint:disable-next-line: no-any
   public parsedContent: Observable<{ [keyName: string]: any }> = this.contentChange.pipe(
+    map((data) => ({
+      ...data,
+      type: this.componentRegister.getComponent(data.type),
```
szenzaro: Move this to `content-viewer.component` changing  `public parsedContent: Observable ...` like you do inside this `map`

ChiaraDipi: It is already in content-viewer.component

PR #21, opened 2020-01-24
Feature/named entities refs

ChiaraDipi:
Add support (parser and component) for references to named entities in the text

### DIFF
--- a/src/app/app.module.ts
+++ b/src/app/app.module.ts
@@ -24,6 +24,7 @@ import { ContentViewerComponent } from './components/content-viewer/content-view
 import { GenericElementComponent } from './components/generic-element/generic-element.component';
 import { GlobalListsComponent } from './components/global-lists/global-lists.component';
 import { NamedEntitiesListComponent } from './components/named-entities-list/named-entities-list.component';
+import { NamedEntityRefComponent } from './components/named-entity-ref/named-entity-ref.component';
 import { NamedEntityDetailComponent } from './components/named-entity/named-entity-detail/named-entity-detail.component';
 import { NamedEntityComponent } from './components/named-entity/named-entity.component';
 import { NoteComponent } from './components/note/note.component';
@@ -78,6 +79,7 @@ export function initializeApp(appConfig: AppConfig) {
     NamedEntitiesListComponent,
     NamedEntityComponent,
     NamedEntityDetailComponent,
+    NamedEntityRefComponent,
     NoteComponent,
     OriginalEncodingViewerComponent,
     OsdComponent,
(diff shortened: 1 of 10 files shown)


src/app/app.module.ts
```diff
@@ -97,6 +98,7 @@ export function initializeApp(appConfig: AppConfig) {
     VersionPanelComponent,
     WitnessPanelComponent,
     XmlBeautifyPipe,
+    NamedEntityRefComponent,
```
szenzaro: Reorder alphabetically

src/app/components/named-entity-ref/named-entity-ref.component.scss
```diff
@@ -0,0 +1,65 @@
+@import "../../../assets/scss/themes";
+@import "../../../assets/scss/mixins";
+@function get-color($key) {
+  @return map-fetch($editionColors, namedEntities $key);
+}
+
+@mixin namedEntityRefColors($color-base, $color-middle, $color-darker) {
+  /* &.location-mainText, &.location-pinned {*/
```
szenzaro: Remove comments

src/app/components/named-entity-ref/named-entity-ref.component.ts
```diff
@@ -0,0 +1,35 @@
...
+  selector: 'evt-named-entity-ref',
+  templateUrl: './named-entity-ref.component.html',
+  styleUrls: ['./named-entity-ref.component.scss'],
+})
+@register
+export class NamedEntityRefComponent {
+  @Input() data: NamedEntityRefData;
+
+  entity$ = this.neParserService.namedEntities$.pipe(
+    share(),
```
szenzaro: ??

src/app/components/named-entity-ref/named-entity-ref.component.ts
```diff
@@ -0,0 +1,35 @@
...
+    map(ne => ne.all.entities.find(e => e.id === this.data.entityId) || 'notFound'),
```
szenzaro: `'notFound'` is not af `NamedEntity` type!

src/app/components/named-entity-ref/named-entity-ref.component.ts
ChiaraDipi: I really don't know why I used it. It's probably a copy-paste from somewhere else. I'll remove it!

src/app/components/named-entity-ref/named-entity-ref.component.ts
ChiaraDipi: I need a way to distinguish among three different statuses: 
- XML Not yet parsed => `undefined`
- XML Parsed and entity found => `NamedEntity` object
- XML Parsed but entity not found => `'notFound'` string

Do you have a better implementation to suggest?

szenzaro: You need to change the entities parser, let's leave it as it is but it should be changed

PR #23, opened 2020-01-28
Feature/named entities occurrences

ChiaraDipi:
I've added the parser to extract NE occurrences based on page list. 
Since everything depends on the page (at least for now), I think this can work. 
I've also handled cases like [Codice Pelavicino](https://pelavicino.labcd.unipi.it/evt) where there could be multiple documents in a single page, and references can be in the same page but belong to different documents.

Navigation to the occurrence page has not been handled in this PR, since navigation is not properly implemented yet.

### DIFF
--- a/src/app/services/xml-parsers/named-entities-parser.service.ts
+++ b/src/app/services/xml-parsers/named-entities-parser.service.ts
@@ -5,7 +5,7 @@ import { map, shareReplay } from 'rxjs/operators';
 import { AppConfig } from '../../app.config';
 import {
   Description, NamedEntities, NamedEntitiesList, NamedEntity, NamedEntityInfo,
-  NamedEntityLabel, NamedEntityOccurrence, NamedEntityType, Relation, XMLElement,
+  NamedEntityLabel, NamedEntityOccurrence, NamedEntityType, PageData, Relation, XMLElement,
 } from '../../models/evt-models';
 import { isNestedInElem, xpath } from '../../utils/dom-utils';
 import { Map } from '../../utils/js-utils';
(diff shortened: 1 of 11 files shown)


src/app/services/xml-parsers/named-entities-parser.service.ts
```diff
@@ -62,17 +64,35 @@ export class NamedEntitiesParserService {
...
+  );
+
   private neListsConfig = AppConfig.evtSettings.edition.namedEntitiesLists || {};
   private tagNamesMap: { [key: string]: string } = {
     persons: 'listPerson',
     places: 'listPlace',
     organizations: 'listOrg',
     relations: 'listRelation',
     events: 'listEvent',
+    occurrences: 'persName[ref], placeName[ref], orgName[ref], geogName[ref], event[ref]',
```
szenzaro: Is this with backticks (`) ?

src/app/services/xml-parsers/named-entities-parser.service.ts
```diff
@@ -167,7 +187,10 @@ export class NamedEntitiesParserService {
       namedEntityType: this.getEntityType(xml.tagName),
       content: Array.from(xml.children).map((subchild: XMLElement) => this.parseEntityInfo(subchild)),
       attributes: this.parseAttributes(xml),
-      occurrences: [],
+      occurrences$: this.occurrences$.pipe(
+        map(occ => occ[elId] || []),
+        shareReplay(1),
```
szenzaro: How many subscriptions?
Putting shareReplay everywhere is not the way ;)

PR #27, opened 2020-02-01
Feature/entities relations

ChiaraDipi:
List of relations is currently inserted among _named entites_ lists, ad for the moments the parser for relations only handle participants from _named entities_ items.

### DIFF
--- a/src/app/components/named-entity-relation/named-entity-relation.component.html
+++ b/src/app/components/named-entity-relation/named-entity-relation.component.html
@@ -17,7 +17,7 @@
             </ng-template>
         </ng-template>
 
-        {{ data.name | lowercase }}
+        {{ data.name | humanize | lowercase }}
 
         <ng-template ngFor let-part let-last="last" [ngForOf]="passiveParts$ | async">
             <ng-template [ngTemplateOutlet]="relationParticipant"
--- a/src/app/components/named-entity-relation/named-entity-relation.component.html
+++ b/src/app/components/named-entity-relation/named-entity-relation.component.html
@@ -32,6 +32,6 @@
     </div>
 </div>
 <ng-template #relationParticipant let-part="part" let-last="last">
-    <span class="relation-entity {{ part.entity?.namedEntityType }}"
+    <span class="relation-entity {{ part.entity?.namedEntityType }}" [class.disabled]="!part.entity"
         (click)="toggleEntity(part.entity)">{{ (part.entity ? part.entity.label : part.id) + (!last ? ', ' : '' ) }}</span>
 </ng-template>
--- a/src/app/components/named-entity-relation/named-entity-relation.component.html
+++ b/src/app/components/named-entity-relation/named-entity-relation.component.html
@@ -1,0 +1,34 @@
+<div class="relation pb-2" [ngClass]="{
+        'p-3': !inEntity,
+        'border-bottom': !inEntity
+    }">
+    <small class="relation-type" *ngIf="data.relationType">{{ data.relationType | lowercase }}</small>
+    <div>
+        <evt-icon [iconInfo]="{icon: 'network-wired', additionalClasses: 'mr-2'}"></evt-icon>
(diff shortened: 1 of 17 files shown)


src/app/components/named-entity-relation/named-entity-relation.component.html
```diff
@@ -0,0 +1,37 @@
...
+
+        <evt-named-entity *ngIf="selectedEntity" [entity]="selectedEntity" [inList]="false"></evt-named-entity>
+    </div>
+    <div class="relation-description text-secondary" *ngIf="data.description.length > 0">
+        <evt-content-viewer *ngFor="let el of data.description" [content]="el"></evt-content-viewer>
+    </div>
+</div>
+<ng-template #relationParticipant let-part="part" let-last="last">
+    <span class="relation-entity {{ part.entity?.namedEntityType }}" [class.disabled]="!part.entity"
+        (click)="toggleEntity(part.entity)">{{ (part.entity ? part.entity.label : part.id) + (!last ? ', ' : '' ) }}</span>
```
szenzaro: Make `(part.entity ? part.entity.label : part.id) + (!last ? ', ' : '' ) ` a getter property?

src/app/components/named-entity-relation/named-entity-relation.component.ts
```diff
@@ -0,0 +1,59 @@
...
+@Component({
+  selector: 'evt-named-entity-relation',
+  templateUrl: './named-entity-relation.component.html',
+  styleUrls: ['./named-entity-relation.component.scss'],
+})
+export class NamedEntityRelationComponent {
+  @Input() data: Relation;
+  @Input() inEntity: boolean;
+
+  // tslint:disable-next-line: variable-name
```
szenzaro: Shouldn't be needed

src/app/components/named-entity-relation/named-entity-relation.component.ts
```diff
@@ -0,0 +1,59 @@
...
+  passiveParts$ = this.getEntities('passiveParts');
+
+  constructor(
+    private neParserService: NamedEntitiesParserService,
+  ) {
+  }
+
+  toggleEntity(entity: NamedEntity) {
+    if (this.inEntity) {
+      // TODO: open entity in list
```
szenzaro: Remove empty block

src/app/components/named-entity-relation/named-entity-relation.component.html
ChiaraDipi: How can I define a getter property with parameters? I do need to check on current value for `last`?

src/app/components/named-entity-relation/named-entity-relation.component.ts
ChiaraDipi: If I remove the comment I get the error `"variable name must be in lowerCamelCase, PascalCase or UPPER_CASE (variable-name)"` since the private property name is `_selectedEntity`. Shall I change the name of the variable?

szenzaro: Do you need the underscore?

src/app/components/named-entity-relation/named-entity-relation.component.ts
```diff
@@ -0,0 +1,65 @@
...
+
+  // tslint:disable-next-line: variable-name
+  private _selectedEntity: NamedEntity;
+
+  set selectedEntity(entity: NamedEntity) {
+    this._selectedEntity = entity;
+  }
+  get selectedEntity() {
+    return this._selectedEntity;
+  }
```
szenzaro: why not using the field directly?

ChiaraDipi: You're totally right!

src/app/components/named-entity-relation/named-entity-relation.component.html
szenzaro: use a function/method

PR #28, opened 2020-02-03
Disable named entity info tab when no details provided for ne

ChiaraDipi:


src/app/components/named-entity/named-entity.component.html
```diff
@@ -16,35 +16,34 @@
...
+                <ng-template ngbTabContent *ngIf="contentOpened">
                     <div class="ne-tab-content">
                         <evt-content-viewer *ngFor="let detail of entity.content" [content]="detail">
                         </evt-content-viewer>
                     </div>
                 </ng-template>
             </ngb-tab>
-            <ngb-tab [title]="'Occurrences' | translate">
-                <ng-template ngbTabContent>
+            <ngb-tab id="occurrences" [title]="'Occurrences' | translate">
```
szenzaro: How does this `id` behave when multiple instances of this component are presenti in the page?

src/app/components/named-entity/named-entity.component.html
```diff
@@ -16,35 +16,34 @@
...
+                <ng-template ngbTabContent *ngIf="contentOpened">
                     <div class="ne-tab-content no-padding">
                         <evt-original-encoding-viewer [originalEncoding]="entity.originalEncoding">
                         </evt-original-encoding-viewer>
                     </div>
                 </ng-template>
             </ngb-tab>
-            <ngb-tab title="EVT JSON Model">
-                <ng-template ngbTabContent>
+            <ngb-tab id="json" title="EVT JSON Model">
```
szenzaro: Same question of the other `id`

src/app/components/named-entity/named-entity.component.html
```diff
@@ -16,35 +16,34 @@
...
+                <ng-template ngbTabContent *ngIf="contentOpened">
                     <div class="ne-tab-content">
                         <span *ngFor="let occurrence of entity.occurrences"> {{ occurrence }} </span>
                         <span *ngIf="entity.occurrences.length === 0">{{'NoOccurrences' | translate}}</span>
                     </div>
                 </ng-template>
             </ngb-tab>
-            <ngb-tab [title]="'XMLSource' | translate">
-                <ng-template ngbTabContent>
+            <ngb-tab id="xml" [title]="'XMLSource' | translate">
```
szenzaro: Same question of the other `id`

src/app/components/named-entity/named-entity.component.html
```diff
@@ -16,35 +16,34 @@
...
     </div>
-    <div class="card-body ne-content" *ngIf="contentOpened">
-        <!-- Tab panes -->
-        <ngb-tabset [justify]="false">
-            <ngb-tab [title]="'Info' | translate">
-                <ng-template ngbTabContent>
+    <div class="card-body ne-content d-none" [class.d-block]="contentOpened">
+        <ngb-tabset [justify]="false" #entityDetails="ngbTabset">
+            <ngb-tab id="info" [title]="'Info' | translate" [disabled]="entity.content.length === 0">
+                <ng-template ngbTabContent *ngIf="contentOpened">
```
szenzaro: careful with `*ngIf` in `ng-template`!
https://angular.io/guide/structural-directives

src/app/components/named-entity/named-entity.component.ts
```diff
@@ -20,6 +23,11 @@ export class NamedEntityComponent implements OnInit {
   toggleContent() {
     if (this.inList) {
       this.contentOpened = !this.contentOpened;
+      if (this.contentOpened) {
+        if (this.entityDetails.activeId === 'info' && this.entity.content.length === 0) {
+          this.entityDetails.select('occurrences');
+        }
+      }
```
szenzaro: This shouldn't be inside `toggleContent`.
Try to make this change as a callback of the event `contentOpened is true`

src/app/components/named-entity/named-entity.component.html
ChiaraDipi: In Bootrap documentation they say that `id` is «The tab identifier. Must be unique for the entire document for proper accessibility support». But it works correctly when multiple instances of this component are present in the page, even if I use the same id for tabs. Shall I use the entity id to create a more appropriate, unique identifier?

src/app/components/named-entity/named-entity.component.html
ChiaraDipi: See previous answer

src/app/components/named-entity/named-entity.component.html
ChiaraDipi: See previous answer

src/app/components/named-entity/named-entity.component.html
szenzaro: It is probably better to not rely on implementation behaviour. Let's put unique ids ;)

ChiaraDipi: Should I generate them programmatically or do you think I can use NamedEntity ID? It should be unique!

szenzaro: since you have the NE-id just use that

PR #31, opened 2020-02-21
Feature/witnesses list extraction

giuliac89:


src/app/models/evt-models.ts
```diff
@@ -105,3 +105,27 @@ export interface NamedEntityRefData extends GenericElementData {
     entityId: string;
     entityType: NamedEntityType;
 }
+
+export interface WitnessesData {
+    witnesses: Map<Witness>;
+    witnessGroups: WitnessGroups;
```
ChiaraDipi: Why didn't you put directly `Map<WitnessGroup>` in here?

src/app/models/evt-models.ts
```diff
@@ -105,3 +105,27 @@ export interface NamedEntityRefData extends GenericElementData {
...
+}
+
+export interface Witness {
+    id: string;
+    name: Description;
```
ChiaraDipi: IMHO you should not use `Description` type in here but directly `GenericElementData[]` or a more specific type. `Description` is intended to be the data model for `<desc>` TEI element and could change in the future if we decide to handle it differently.

src/app/models/evt-models.ts
```diff
@@ -105,3 +105,27 @@ export interface NamedEntityRefData extends GenericElementData {
...
+    attributes: AttributesData;
+    content: Description;
```
ChiaraDipi: See previous comment (l. 116)

src/app/models/evt-models.ts
```diff
@@ -105,3 +105,27 @@ export interface NamedEntityRefData extends GenericElementData {
...
+}
+
+export interface WitnessGroups {
+    groups: Map<WitnessGroup>;
+}
+export interface WitnessGroup {
+    id: string;
+    name: string;
+    attributes: AttributesData;
+    content: string[];
```
ChiaraDipi: If this is intended to list the ids of witnesses, I would change the property name in something like `witnesses` or `witnessesIds`. If not, ignore my comment.

src/app/models/evt-models.ts
```diff
@@ -105,3 +105,27 @@ export interface NamedEntityRefData extends GenericElementData {
...
+    content: Description;
+    group: string;
+}
+
+export interface WitnessGroups {
+    groups: Map<WitnessGroup>;
+}
+export interface WitnessGroup {
+    id: string;
+    name: string;
```
ChiaraDipi: Are you sure that this is a simple string and not a `GenericElementData[]` as for the `Witness`?

src/app/models/evt-models.ts
```diff
@@ -105,3 +105,27 @@ export interface NamedEntityRefData extends GenericElementData {
...
+    attributes: AttributesData;
+    content: string[];
+    group: string;
```
ChiaraDipi: If this is intended to list the ids of nested groups, I would change the property name in something like `groups` or `groupsIds`. If not, I don't really get what this property should represent.

src/app/services/xml-parsers/witnesses-parser.service.ts
```diff
@@ -0,0 +1,134 @@
...
+})
+export class WitnessesParserService {
+  public readonly parsedList$ = this.editionDataService.parsedEditionSource$.pipe(
+    map((source) => this.parseWitnessesData(source)),
+    shareReplay(1),
+  );
+
+  private witListTagName = 'listWit';
+  private witTagName = 'witness';
+  private witNameAttr = 'type=\'siglum\'';
```
ChiaraDipi: Use backticks in order to avoid the back slashes: ``` `type='siglum'` ```, or just use double quotation marks inside the single ones `'type="siglum"'`.

src/app/services/xml-parsers/witnesses-parser.service.ts
```diff
@@ -0,0 +1,134 @@
...
+      id: wit.getAttribute('xml:id') || xpath(wit),
+      name: this.parseWitnessName(wit) || wit.getAttribute('xml:id') || xpath(wit),
+      attributes: this.genericParserService.parseAttributes(wit),
+      content: this.parseWitnessContent(wit),
+      group: this.parseParentGroupId(wit),
+    };
+  }
+
+  private parseWitnessName(wit: XMLElement): Description {
+    const witNameEl = wit.querySelector<XMLElement>('*[' + this.witNameAttr + ']');
```
ChiaraDipi: You can use the syntax ``` `*[${this.withNameAttr}]` ``` (with backticks)

src/app/services/xml-parsers/witnesses-parser.service.ts
```diff
@@ -0,0 +1,134 @@
...
+
+  private parseWitnessContent(wit: XMLElement): Description {
+    return Array.from(wit.childNodes)
+      .filter((child) => child.nodeName !== this.witListTagName && child.textContent.trim().length !== 0)
+      .map((child: XMLElement) => this.genericParserService.parse(child));
+  }
+
+  private parseWitnessesGroups(lists: XMLElement[]): WitnessGroups {
+    return {
+      groups: this.parseGroups(lists),
```
ChiaraDipi: Why not directly `return this.parseGroups(lists);` ?

src/app/services/xml-parsers/witnesses-parser.service.ts
```diff
@@ -0,0 +1,134 @@
...
+
+  private parseGroups(lists: XMLElement[]) {
+    return lists.filter((list) => isNestedInElem(list, list.tagName))
+      .map((list) => this.parseWitnessGroup(list))
+      .reduce((x, y) => x.concat(y), [])
+      .reduce((list, item) => {
+        list[item.id] = item;
+
+        return list;
+      },      {});
```
ChiaraDipi: You can simplify with:
```
.reduce((list, item) => ({
  ...list,
  [item.id]: item,
}), {})
```

src/app/services/xml-parsers/witnesses-parser.service.ts
```diff
@@ -0,0 +1,134 @@
...
+
+  private parseWitnessesList(lists: XMLElement[]) {
+    return lists.filter((list) => !isNestedInElem(list, list.tagName))
+      .map((list) => this.parseWitnesses(list))
+      .reduce((x, y) => x.concat(y), [])
+      .reduce((list, item) => {
+        list[item.id] = item;
+
+        return list;
+      },      {});
```
ChiaraDipi: You can simplify this:
```
.reduce((list, item) => ({
  ...list,
  [item.id]: item,
}), {})
```

src/app/services/xml-parsers/witnesses-parser.service.ts
```diff
@@ -0,0 +1,134 @@
...
+  private parseGroupContent(list: XMLElement) {
+    return Array.from(list.children)
+      .filter(({nodeName}) => nodeName === this.witListTagName || nodeName === this.witTagName)
+      .map((child) => child.getAttribute('xml:id'));
+  }
+
+  private parseParentGroupId(element: XMLElement) {
+    let parentEl = element.parentElement;
+
+    if (parentEl) {
```
szenzaro: What happens is `parentEl` is `undefined`?

giuliac89: This line is unnecessary and can be removed. parentEl will never be undefined.

src/app/services/xml-parsers/witnesses-parser.service.ts
```diff
@@ -0,0 +1,118 @@
...
+      content: this.parseWitnessContent(wit),
+      groupId: this.parseParentGroupId(wit),
+    };
+  }
+
+  private parseWitnessName(wit: XMLElement) {
+     // TODO use ‘?’ operator after update tu angular 9
+    const witNameEl = wit.querySelector<XMLElement>(`*[${this.witNameAttr}]`);
+
+    if (witNameEl) {
```
szenzaro: What do you return if witNameEl is undefined?

src/app/models/evt-models.ts
ChiaraDipi: I saw that it can contain also a `<head>` or `<desc>`, thus it may be more complex than a string

ChiaraDipi: Nevermind, it is probably enough to have simply string for now

src/app/services/xml-parsers/witnesses-parser.service.ts
```diff
@@ -0,0 +1,120 @@
...
+      id: wit.getAttribute('xml:id') || xpath(wit),
+      name: this.parseWitnessName(wit) || wit.getAttribute('xml:id') || xpath(wit),
+      attributes: this.genericParserService.parseAttributes(wit),
+      content: this.parseWitnessContent(wit),
+      groupId: this.parseParentGroupId(wit),
+    };
+  }
+
+  private parseWitnessName(wit: XMLElement) {
+    // TODO use ‘?’ operator after update tu angular 9
```
ChiaraDipi: Syntax error in comment "tu" => "to"

PR #38, opened 2020-03-04
Feature/edition title

ChiaraDipi:


src/app/main-header/main-header.component.html
```diff
@@ -1,7 +1,7 @@
 <nav class="navbar justify-content-between temp-header">
     <a class="navbar-brand">
-        <img src="../../assets/images/logo.png" height="30" class="d-inline-block align-top" alt="">
-        <span >{{ editionConfig?.title || 'defaultTitle' | translate }}</span>
+        <img src="../../assets/images/logo.png" height="30" class="d-inline-block align-top mr-2" alt="">
```
szenzaro: this path will probably fail with a different base href

src/app/services/xml-parsers/prefatory-matter-parser.service.ts
```diff
@@ -0,0 +1,26 @@
...
+import { EditionDataService } from '../edition-data.service';
+
+@Injectable({
+  providedIn: 'root',
+})
+export class PrefatoryMatterParserService {
+  public readonly title$ = this.editionDataService.parsedEditionSource$.pipe(
+    map((source) => this.parseEditionTitle(source)),
+    shareReplay(1),
+    tap(t => console.log(t)),
```
szenzaro: no `console.log`s

ChiaraDipi: Ups. I forgot to remove it.

PR #42, opened 2020-03-06
Add link to edition home on edition title if editionHome set in config

ChiaraDipi:


src/app/main-header/main-header.component.scss
```diff
@@ -63,4 +63,8 @@
     }
     font-size: .9rem;
     vertical-align: text-bottom;
+}
+
+.pointer {
```
szenzaro: This could be useful elsewhere. Maybe it is worth to put it in the global style

PR #46, opened 2020-03-10
Feature/edition level selector in text panel

ChiaraDipi:


src/app/view-modes/collation/collation.component.html
```diff
@@ -1,6 +1,9 @@
 <gridster [options]="options">
     <gridster-item [item]="textPanelItem">
-        <evt-text-panel></evt-text-panel>
+        <evt-text-panel
+            [editionLevelID]="currentEditionLevel | async"
+            (editionLevelChange)="handleEditionLevelChange($event)"
+        ></evt-text-panel>
```
giuliac89: Is this selector necessary in the collation view mode? I think it's not necessary in any case

ChiaraDipi: Ok. Maybe we should just set the param directly as "critical" in this case, what do you think?

PR #47, opened 2020-03-15
Feature/apparatus entries extraction

giuliac89:


src/app/models/evt-models.ts
```diff
@@ -127,6 +127,23 @@ export interface WitnessGroup {
     groupId: string;
 }
 
+export type ApparatusEntriesList = Map<ApparatusEntry>;
+
+export interface ApparatusEntry extends GenericElementData {
+    id: string;
+    content: ApparatusReading[];
+    note: NoteData;
```
ChiaraDipi: Does TEI allow to have multiple `<note>`s within a single `<app>`? If so, we should maybe anticipate this particular encoding and handle `note` as an array of `NoteData`

src/app/models/evt-models.ts
```diff
@@ -127,6 +127,23 @@ export interface WitnessGroup {
...
+    variance: number;
+    originalEncoding: string;
+}
+
+export interface ApparatusReading extends GenericElementData {
+    tagname: string;
```
ChiaraDipi: Why do you need to store tag name information?

src/app/services/xml-parsers/apparatus-entries-parser.service.ts
```diff
@@ -0,0 +1,128 @@
...
+  }
+
+  private parseAppEntriesList(document: XMLElement): ApparatusEntriesList {
+    const appEntries = Array.from(document.querySelectorAll<XMLElement>(this.appEntryTagName));
+
+    return this.parseAppEntries(appEntries);
+  }
+
+  private parseAppEntries(appEntries: XMLElement[]) {
+    this.witParserService.witnesses$.subscribe(data => this.parsedWitnesses = data);
```
ChiaraDipi: What happens if `this.witParserService.witnesses$` does not resolve immediately? If you need parsed witnesses data to calculate variance, you should be sure that those information are ready when you use them. So you can either put the parsing of app entry inside the subscription, or handle variance as an observable that wait `this.witParserService.witnesses$` to be resolved. I probably prefer the second option

src/app/services/xml-parsers/apparatus-entries-parser.service.ts
```diff
@@ -0,0 +1,128 @@
...
+
+  private parseAppEntry(appEntry: XMLElement): ApparatusEntry {
+    this.significantRdgNumber = 0;
+
+    return {
+      type: 'ApparatusEntryComponent',
+      id: appEntry.getAttribute('xml:id') || xpath(appEntry),
+      attributes: this.genericParserService.parseAttributes(appEntry),
+      content: this.parseAppReadings(appEntry),
+      note: this.genericParserService.parse(appEntry.querySelector<XMLElement>(this.noteTagName)),
```
ChiaraDipi: See previous comment: what happen when there are more `<note>` elements?

src/app/services/xml-parsers/apparatus-entries-parser.service.ts
```diff
@@ -0,0 +1,128 @@
...
+            attributes: {},
+            content: [],
+          };
+        }
+
+        return this.genericParserService.parse(child);
+      });
+  }
+
+  private parseReadingSignificance(rdg: XMLElement) {
```
ChiaraDipi: I would organize differently the two functions `parseReadingSignificance ` and `isSignificantReading `. The second one should just check if it is significant and return the boolean. I would probably put the other computations inside `parseReadingSignificance` since its name suggests that besides checking if is significant you also do some other stuff with that info.

src/app/services/xml-parsers/apparatus-entries-parser.service.ts
```diff
@@ -0,0 +1,128 @@
...
+    return rdg.parentElement.tagName === this.readingGroupTagName ?
```
ChiaraDipi: What if you have consider not significant those readings with `@type="orthographic"` and you have this encoding:
```
<rdgGrp resp="m">
   <rdg type="orthographic">Loremm</rdg>
   <rdg>Irem</rdg>
</rdgGrp>
```
in this case, when parsing the first `rdg`, since parent element is a `<rdgGrp>`, it will run the significance check on parent element attributes only and the reading will be marked as significant. You should probably also check whether the `<rdg>` as attributes or not, and if it has, check those first (then check parent's if the reading results to be significant).

src/app/services/xml-parsers/apparatus-entries-parser.service.ts
```diff
@@ -0,0 +1,128 @@
...
+      this.isSignificantReading(rdg.parentElement.attributes) :
+      this.isSignificantReading(rdg.attributes);
+  }
+
+  private isSignificantReading(attributes: NamedNodeMap) {
+    const notSignificantReading = AppConfig.evtSettings.edition.notSignificantVariant;
+    let isSignificant = true;
+
+    Array.from(attributes).forEach((attr) => {
```
ChiaraDipi: You probably don't need to loop over all attributes, but stop once `isSignificant` changes to `false`. Also, if `notSignificantReading` is not defined you loop over every attribute for nothing.

src/app/services/xml-parsers/apparatus-entries-parser.service.ts
```diff
@@ -0,0 +1,128 @@
...
+
+    if (isSignificant) {
+      this.significantRdgNumber++;
+    }
+
+    return isSignificant;
+  }
+
+  private getVariance() {
+    return this.significantRdgNumber / Object.keys(this.parsedWitnesses).length;
```
ChiaraDipi: For each apparatus entry, variance is calculated as the number of important reading on the number of witnesses. But if you use this "global" variable (`significantRdgNumber`) this information is kinda lost because you have to go back to the `parseAppEntry` function to see that it is reset to zero. I would probably wait for each app children (lemma and readings) to be parsed and then calculate the variance from that reasults, even if it means to have a double loop over children.

src/app/models/evt-models.ts
```diff
@@ -127,6 +127,23 @@ export interface WitnessGroup {
     groupId: string;
 }
 
+export type ApparatusEntriesList = Map<ApparatusEntry>;
+
+export interface ApparatusEntry extends GenericElementData {
+    id: string;
+    content: ApparatusReading[];
```
ChiaraDipi: I would distinguish lemma from other readings

src/app/models/evt-models.ts
giuliac89: I need it to distinguish the lemma from other readings when they are displayed in the component. Do you think there is another way?

src/app/models/evt-models.ts
giuliac89: I think it's not necessary, because lem and rdg are all readings.  I made this distinction through the `tagname` property in the ApparatusReading interface.

src/app/services/xml-parsers/apparatus-entries-parser.service.ts
giuliac89: Done

src/app/services/xml-parsers/apparatus-entries-parser.service.ts
giuliac89: Done

src/app/services/xml-parsers/apparatus-entries-parser.service.ts
giuliac89: Done

src/app/services/xml-parsers/apparatus-entries-parser.service.ts
giuliac89: Done

src/app/models/evt-models.ts
giuliac89: Done

src/app/models/evt-models.ts
```diff
@@ -127,6 +127,23 @@ export interface WitnessGroup {
...
+    note: NoteData[];
+    variance: number;
+    originalEncoding: string;
+}
+
+export interface ApparatusReading extends GenericElementData {
+    tagname: string;
+    id: string;
+    wits: string[];
```
szenzaro: Shouldn't these be part of the lemma/rdg?

src/app/services/xml-parsers/apparatus-entries-parser.service.ts
```diff
@@ -0,0 +1,131 @@
...
+    private genericParserService: GenericParserService,
+    private witParserService: WitnessesParserService,
+  ) {
+  }
+
+  private parseAppEntriesList(document: XMLElement): ApparatusEntriesList {
+    const appEntries = Array.from(document.querySelectorAll<XMLElement>(this.appEntryTagName));
+
+    return this.parseAppEntries(appEntries);
+  }
```
szenzaro: I think we should generalise this kind of function to be used everywhere in the parsers.
@ChiaraDipi

src/app/services/xml-parsers/apparatus-entries-parser.service.ts
szenzaro: we should try a declarative approach leveraging rxjs

src/app/services/xml-parsers/apparatus-entries-parser.service.ts
```diff
@@ -0,0 +1,131 @@
...
+
+  private parseAppEntries(appEntries: XMLElement[]) {
+    this.witParserService.witnesses$.subscribe(data => this.parsedWitnesses = data);
+    const parsedEntries = appEntries.map((appEntry) => this.parseAppEntry(appEntry));
+
+    return arrayToMap(parsedEntries, 'id');
+  }
+
+  private parseAppEntry(appEntry: XMLElement): ApparatusEntry {
+    this.significantRdgNumber = 0;
```
szenzaro: ???

src/app/services/xml-parsers/apparatus-entries-parser.service.ts
szenzaro: What's the meaning of "significant" reading?

src/app/services/xml-parsers/apparatus-entries-parser.service.ts
```diff
@@ -0,0 +1,131 @@
...
+    return isSignificant;
+  }
+
+  private isSignificantReading(notSignificantReading: string[], attributes: NamedNodeMap) {
+    return !Array.from(attributes).some(({name, value}) => {
+      return notSignificantReading.includes(`${name} =  ${value}`);
+    });
+  }
+
+  private parseAppNotes(appEntry: XMLElement) {
```
szenzaro: Can the notes parse be used instead?

src/app/services/xml-parsers/apparatus-entries-parser.service.ts
szenzaro: could the variance "factor" be a part of the apparatus/reading model?

src/app/models/evt-models.ts
giuliac89: @ChiaraDipi @szenzaro Is it ok if I create an ApparatusLemma interface for lemma and an ApparatusReading interface for other readings? Or maybe it's better to group everything in "ApparatusReading" and create new fields? I didn't understand how you imagined it.

src/app/services/xml-parsers/apparatus-entries-parser.service.ts
giuliac89: Do you mean the parser in generic-parser? If so, no, because that parser takes an XMLElement as parameter and not an array.

src/app/services/xml-parsers/apparatus-entries-parser.service.ts
giuliac89: It's already a field of the ApparatusEntry model

src/app/models/evt-models.ts
giuliac89: Yes, if we distinguish them.

src/app/services/xml-parsers/apparatus-entries-parser.service.ts
giuliac89: What do you mean? Do you refer to a general parseList function?

src/app/services/xml-parsers/apparatus-entries-parser.service.ts
giuliac89: "Significant reading" is in contrast with "non-significant reading", that is, a reading that does not present a particular variance with respect to the lemma. For example an ortographic variation is not as important as a lexicographic variation. We recover the "non-significant reading" from the config file.

src/app/services/xml-parsers/apparatus-entries-parser.service.ts
szenzaro: Not in this file it seems :P

src/app/models/evt-models.ts
szenzaro: You can just call the interfaces "Lemma" and "Reading" ;)

src/app/models/evt-models.ts
szenzaro: https://www.google.com/url?sa=i&url=https%3A%2F%2Fgiphy.com%2Fexplore%2Fben-lucas&psig=AOvVaw1KL2QpPzsJUXSDea9OBxXY&ust=1585832606753000&source=images&cd=vfe&ved=0CAIQjRxqFwoTCJC3xPykx-gCFQAAAAAdAAAAABAD

src/app/services/xml-parsers/apparatus-entries-parser.service.ts
szenzaro: Yes, but maybe in a refactoring task

src/app/services/xml-parsers/apparatus-entries-parser.service.ts
ChiaraDipi: In traditional printed critical edition non-significant reading where usually excluded in order to save space. In digital edition they can now be encoded and visualized since there are no space problems anymore; but in a situation with a very high number of witnesses and a very high variance is sometimes preferable to hide them. That's why we've decided to distinguish them.

src/app/models/evt-models.ts
giuliac89: Done

src/app/models/evt-models.ts
giuliac89: Done

src/app/app.config.ts
```diff
@@ -77,6 +77,7 @@ export interface EditionConfig {
         relations: NamedEntitiesListsConfig;
         events: NamedEntitiesListsConfig;
     }>;
+    notSignificantVariant: string[];
```
szenzaro, line 80: `notSignificantVariantS`

src/app/services/xml-parsers/apparatus-entries-parser.service.ts
```diff
@@ -0,0 +1,152 @@
...
+      variance: 0,
+      originalEncoding: getOuterHTML(appEntry),
+    };
+
+    entry.variance = this.calcVariance(entry.content, witsNumber);
+
+    return entry;
+  }
+
+  private parseAppReadings(appEntry: XMLElement): Array<Lemma | Reading> {
```
szenzaro, line 68: So lemma is considered a reading?

src/app/services/xml-parsers/apparatus-entries-parser.service.ts
```diff
@@ -0,0 +1,152 @@
...
+  }
+
+  private parseAppEntry(appEntry: XMLElement, witsNumber: number): ApparatusEntry {
+    const entry = {
+      type: 'ApparatusEntryComponent',
+      id: appEntry.getAttribute('xml:id') || xpath(appEntry),
+      attributes: this.genericParserService.parseAttributes(appEntry),
+      content: this.parseAppReadings(appEntry),
+      note: this.parseAppNotes(appEntry),
+      variance: 0,
```
szenzaro, line 59: make the variance an observable instead of a plain value.

src/app/services/xml-parsers/apparatus-entries-parser.service.ts
giuliac89: Yes, the lemma is a reading. In the TEI Guidelines:

> the term lemma is used here in the text-critical sense of ‘the reading accepted as that of the original or of the base text’.

Right @ChiaraDipi?

src/app/models/evt-models.ts
```diff
@@ -127,6 +127,28 @@ export interface WitnessGroup {
...
+
+export interface Lemma extends GenericElementData {
+    id: string;
+    wits: string[];
+    significant: boolean;
+}
+
+export interface Reading extends GenericElementData {
+    id: string;
+    wits: string[];
```
szenzaro, line 148: Are this `witIDs`?

src/app/models/evt-models.ts
```diff
@@ -127,6 +127,28 @@ export interface WitnessGroup {
...
+    id: string;
+    content: Array<Lemma | Reading>;
+    notes: NoteData[];
+    variance: number;
+    originalEncoding: string;
+}
+
+export interface Lemma extends GenericElementData {
+    id: string;
+    wits: string[];
```
szenzaro, line 142: Are this `witIDs`?

src/app/models/evt-models.ts
giuliac89: yes

src/app/models/evt-models.ts
```diff
@@ -127,6 +127,28 @@ export interface WitnessGroup {
     groupId: string;
 }
 
+export type ApparatusEntriesList = Map<ApparatusEntry>;
```
szenzaro, line 130: Just use 'Map<ApparatusEntry>', also because it is not a List

src/app/services/xml-parsers/apparatus-entries-parser.service.ts
```diff
@@ -0,0 +1,153 @@
...
+
+  private parseReadingSignificance(rdg: XMLElement) {
+    const notSignificantReadings = AppConfig.evtSettings.edition.notSignificantVariants;
+    let isSignificant = true;
+
+    if (notSignificantReadings.length > 0) {
+      isSignificant = this.isSignificantReading(notSignificantReadings, rdg.attributes);
+      if (isSignificant && rdg.parentElement.tagName === this.readingGroupTagName) {
+        isSignificant = this.isSignificantReading(notSignificantReadings, rdg.parentElement.attributes);
+      }
```
szenzaro, line 126:
```
return notSignificantReadings.length === 0 || 
this.isSignificantReading(notSignificantReadings, rdg.attributes) || 
(rdg.parentElement.tagName === this.readingGroupTagName && this.isSignificantReading(notSignificantReadings, rdg.parentElement.attributes))
```

src/app/services/xml-parsers/apparatus-entries-parser.service.ts
```diff
@@ -0,0 +1,153 @@
...
+            attributes: {},
+            content: [],
+          };
+        }
+
+        return this.genericParserService.parse(child);
+      });
+  }
+
+  private parseReadingSignificance(rdg: XMLElement) {
```
szenzaro, line 118: This function returns a boolean,  is it still a "parse" function?
Would an alternative name like `readingIsSignificant` fit better?

src/app/services/xml-parsers/apparatus-entries-parser.service.ts
giuliac89: @szenzaro I think this solution it's not so clear... I try it but it doesn't work!

src/app/models/evt-models.ts
giuliac89: ?

src/app/models/evt-models.ts
```diff
@@ -127,6 +127,28 @@ export interface WitnessGroup {
...
+
+export interface ApparatusEntry extends GenericElementData {
+    id: string;
+    content: Array<Lemma | Reading>;
+    notes: NoteData[];
+    variance: number;
+    originalEncoding: string;
+}
+
+export interface Lemma extends GenericElementData {
```
szenzaro, line 140: Use type alias `type Lemma = Reading` since they are the same

PR #48, opened 2020-04-04
Feature/critical comment notes

ChiaraDipi:


src/app/models/parsed-elements.ts
```diff
@@ -15,6 +15,10 @@ export type HTMLData = GenericElementData & {
 export interface TextData extends GenericElementData {
     text: string;
 }
+export type NoteType = 'critical' | 'comment';
```
szenzaro, line 18: Are these fixed or may change by configuration?

src/app/services/xml-parsers/generic-parser.service.ts
```diff
@@ -70,15 +70,19 @@ export class GenericParserService {
     return genericElement;
   }
 
-  private parseNote(xml: XMLElement): NoteData {
+  private parseNote(xml: XMLElement): NoteData | GenericElementData {
```
szenzaro, line 73: When does it returns a `GenericElementData`?

ChiaraDipi: Footer notes and note nested in named entity are handled as generic elements, because they should not be rendered with the exponent and the popover. 
But, now that you make me think about it, it may be better to put this information inside the NoteComponent and distinguish the layout there. What do you think?

src/app/models/parsed-elements.ts
ChiaraDipi: They're fixed

src/app/services/xml-parsers/generic-parser.service.ts
szenzaro: I agree

src/app/models/parsed-elements.ts
ChiaraDipi: If we move the layout logic inside the note component (to output differently notes in footer or in entities) this dual distinction between critical and comment is no more valid. Moreover if we have a look at the TEI guidelines there could be assigned with a very generic string.

src/assets/scss/_themes.scss
```diff
@@ -43,6 +43,8 @@ $editionColors: (
         eventMiddle: #ffffb0,
         eventDarker: #fcfc60,
     ),
+    commentNotes: rgb(87, 14, 105),
+    criticalNotes: rgb(24, 70, 155),
```
giuliac89, line 47: In this file I think it's better to use the variables for the colors of the theme, but we can refactor it in the future.

PR #54, opened 2020-04-15
Add parser for stand off notes

ChiaraDipi:
The parser works for `<ptr>` elements with `@type="noteAnchor"` and have a `@target` pointing to a `<note>` defined anywhere in the xml.

### DIFF
--- a/src/app/services/xml-parsers/generic-parser.service.ts
+++ b/src/app/services/xml-parsers/generic-parser.service.ts
@@ -13,7 +13,7 @@ function complexElements(nodes: NodeListOf<ChildNode>): ChildNode[] {
   return Array.from(nodes).filter((n) => n.nodeType !== 8);
 }
 
-type SupportedTagNames = 'event' | 'geogname' | 'note' | 'orgname' | 'persname' | 'placename';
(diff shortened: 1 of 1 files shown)


src/app/services/xml-parsers/generic-parser.service.ts
```diff
@@ -121,6 +122,18 @@ export class GenericParserService {
     };
   }
 
+  private parsePtr(xml: XMLElement) {
+    if (xml.getAttribute('type') === 'noteAnchor' && xml.getAttribute('target')) {
+      const noteId = xml.getAttribute('target').replace('#', '');
```
szenzaro, line 127: change `replace` with `slice`?

ChiaraDipi: I prefer `replace`, because if someone forget to use the "#" before the referenced id, the transformation still works.

src/app/services/xml-parsers/generic-parser.service.ts
```diff
@@ -1,5 +1,5 @@
 import { Injectable } from '@angular/core';
-import { Subject } from 'rxjs';
+import { Subject, from } from 'rxjs';
```
giuliac89, line 2: I think this value is never read. Check it!

PR #56, opened 2020-04-19
Feature/paragraphs viz

PaoloVallesi:


src/app/components/paragraph/paragraph.component.html
```diff
@@ -0,0 +1,7 @@
+<p>
+	<span class="p-num">{{data.n}}.</span>
+	<evt-content-viewer class="genericElement {{data.class}}" *ngFor="let el of data.content" [content]="el"
```
giuliac89, line 3: Why the "genericElement" class?

src/app/services/xml-parsers/generic-parser.service.ts
```diff
@@ -60,6 +61,18 @@ export class GenericParserService {
...
+      type: 'ParagraphComponent',
+      class: xml.tagName ? xml.tagName.toLowerCase() : '',
+      content: this.parseChildren(xml),
+      attributes: this.parseAttributes(xml),
+      n: this.parseAttributes(xml).n || '-1',
+    };
+
+    return paragraphComponent;
+  }
+
```
giuliac89, line 75: What about managing a paragraph with a `<pb>` element inside?
For example:
```
<div>
   <pb xml:id="page1" />
   <p n="2">
        Lorem ipsum...
   <p>
   <p n="3">
       Ispum dolor  <pb xml:id="page2"/> sit amet sit amet sit amet....
   </p>
   <p n="4">
      [...]
   </p>
</div>
```

src/app/components/paragraph/paragraph.component.scss
```diff
@@ -0,0 +1,9 @@
+@import "../../../assets/scss/themes";
+
+.p-num {
+  font-size: 1rem;
+  padding-right: .3rem;
+  @include themify($themes){
```
giuliac89, line 6: Missing space after `@include themify($themes)`

src/app/services/xml-parsers/generic-parser.service.ts
```diff
@@ -60,6 +61,18 @@ export class GenericParserService {
     return text;
   }
 
+  public parseParagrah(xml: XMLElement): ParagraphData {
+    const paragraphComponent: ParagraphData = {
+      type: 'ParagraphComponent',
+      class: xml.tagName ? xml.tagName.toLowerCase() : '',
```
giuliac89, line 67: Why this condition? A XMLElement has always a tagName

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Updated to Angular 9
 
 ### Added
+- Numbered pharagraphs
```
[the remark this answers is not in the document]
szenzaro: No, keep inversed chronological order

src/app/models/parsed-elements.ts
```diff
@@ -22,4 +22,8 @@ export type NoteData = GenericElementData & {
     exponent: string;
 };
 
+export interface ParagraphData extends GenericElementData {
+    n: string;
```
szenzaro, line 26: What does `n` represents?

src/app/services/xml-parsers/generic-parser.service.ts
szenzaro: `pb`s would be managed recursively by the parser

src/app/models/parsed-elements.ts
ChiaraDipi: `n` is the number of the paragraph. It can be defined in the xml, or it can be automatically created. For the moment we need to handle only the first case.

src/app/services/xml-parsers/generic-parser.service.ts
```diff
@@ -13,7 +13,7 @@ function complexElements(nodes: NodeListOf<ChildNode>): ChildNode[] {
   return Array.from(nodes).filter((n) => n.nodeType !== 8);
 }
 
-type SupportedTagNames = 'event' | 'geogname' | 'note' | 'orgname' | 'persname' | 'placename' | 'ptr';
+type SupportedTagNames = 'event' | 'geogname' | 'note' | 'orgname' | 'persname' | 'placename' | 'ptr' | 'p';
```
ChiaraDipi, line 16: Shall we order alphabetically also this list?

src/app/services/xml-parsers/generic-parser.service.ts
ChiaraDipi: I agree. There will be a specific parser for `<pb>`s, but is a subject for a new PR!

src/app/services/xml-parsers/generic-parser.service.ts
PaoloVallesi: Copy-Paste error... thinking about it, does it actually make sense having a class attribute here?

src/app/components/paragraph/paragraph.component.html
PaoloVallesi: Copy-Paste error from generic-component in here too! 🙄
Also, i'm going to move the class attribute since `{{data.class}}` refers to the `<p>` tagname.

ChiaraDipi: We can probably drop the class attribute in this case, since it will not make much sense to have `<p class="p">` (knowing that the class is the tag name of the original `<p>` in the xml)

src/app/services/xml-parsers/generic-parser.service.ts
ChiaraDipi: No, in this case having the class `p` for the paragraph component is no actually useful. You can remove it!

src/app/services/xml-parsers/generic-parser.service.ts
giuliac89: Maybe it's better, but can we do it in another PR?

giuliac89: Maybe it's better!

ChiaraDipi: Yeah, let's do it!

PR #57, opened 2020-05-01
Feature/ed level susceptible mixin

ChiaraDipi:
Add support for edition level susceptible components which have different layouts depending on selected edition level. This solution uses mixin, that's why there was the need to switch from base component to mixin also for HighlightableComponent.


src/app/components/components-mixins.ts
```diff
@@ -0,0 +1,15 @@
+import { Input } from '@angular/core';
+import { EditionLevelType } from '../app.config';
+import { EntitiesSelectItem } from './entities-select/entities-select.component';
+
+// tslint:disable-next-line: no-any
+export type Constructor<T> = new (...args: any[]) => T;
```
szenzaro, line 6: Where do you use this?

ChiaraDipi: Right, I forgot to remove it

src/app/components/content-viewer/content-viewer.component.ts
```diff
@@ -50,13 +59,15 @@ export class ContentViewerComponent implements OnDestroy {
   public inputs: Observable<{ [keyName: string]: any }> = combineLatest([
     this.contentChange,
     this.itemsToHighlightChange,
+    this.editionLevelChange,
   ]).pipe(
-    map(([data, ith]) => {
+    map(([data, ith, el]) => {
```
szenzaro, line 64: change to `[data, itemsToHighlight, editionLevel]`?

PR #60, opened 2020-05-02
Add FacsimileParser service to handle surface and zone parsing

ChiaraDipi:


src/app/services/xml-parsers/facsimile-parser.service.ts
```diff
@@ -0,0 +1,117 @@
...
+})
+export class FacsimileParserService {
+
+    constructor(
+        private genericParser: GenericParserService,
+    ) {
+    }
+
+    parseSurfaces(xml: XMLElement): Surface[] {
+        if (xml) {
```
szenzaro, line 59: It is better to return an empty array instead of an implicit undefined.

```
if (!xml) { return [] }
return Array.from( ... )
```

src/app/services/xml-parsers/facsimile-parser.service.ts
```diff
@@ -0,0 +1,117 @@
...
+                        lines: zones.filter(z => z.rendition === 'Line') as ZoneLine[],
+                        hotspots: zones.filter(z => z.rendition === 'HotSpot') as ZoneHotSpot[],
+                    },
+                };
+            });
+        }
+    }
+
+    parseZones(xml: XMLElement): Zone[] {
+        if (xml) {
```
szenzaro, line 77: Avoid implicit `undefined`

src/app/services/xml-parsers/facsimile-parser.service.ts
```diff
@@ -0,0 +1,117 @@
...
+                    rendition: attributes.rendition as ZoneRendition,
+                    rotate: attributes.rotate ? parseInt(attributes.rotate, 10) || 0 : 0,
+                    content: this.genericParser.parseChildren(z),
+                };
+            });
+        }
+    }
+
+    parseGraphics(xml: XMLElement): Graphic[] {
+        if (xml) {
```
szenzaro, line 109: Avoid implicit undefined

src/app/services/xml-parsers/facsimile-parser.service.ts
```diff
@@ -0,0 +1,117 @@
...
+import { GenericParserService, ParsedElement } from './generic-parser.service';
+
+export type ZoneRendition = 'Line' | 'HotSpot'; // EVT rule to distinguish lines for ITL from HotSpots
+export type ZoneCoords = {
+    ulx: number;
+    uly: number;
+    lrx: number;
+    lry: number;
+} | {
+    points: string;
```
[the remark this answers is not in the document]
ChiaraDipi: I've changed approach in order to have always the same type for coordinates. Check it

src/app/services/xml-parsers/facsimile-parser.ts
```diff
@@ -0,0 +1,91 @@
...
+                    };
+                });
+        } else {
+            const ul: Point = {
+                x: attributes.ulx ? parseFloat(attributes.ulx) : undefined,
+                y: attributes.uly ? parseFloat(attributes.uly) : undefined,
+            };
+            const lr: Point = {
+                x: attributes.lrx ? parseFloat(attributes.lrx) : undefined,
+                y: attributes.lry ? parseFloat(attributes.lry) : undefined,
```
szenzaro, line 27: simplify this patter using `parseFloat(x) || undefined`

PR #61, opened 2020-05-02
Feature/line breaks

ChiaraDipi:


src/app/models/parsed-elements.ts
```diff
@@ -26,4 +26,11 @@ export interface ParagraphData extends GenericElementData {
     n: string;
 }
 
+export interface LbData extends GenericElementData {
+    id: string;
+    n?: string;
+    facs?: string; // Needed to handle ITL
+    rend?: string;
```
giuliac89, line 33: An example with the rend attribute?

PR #64, opened 2020-05-17
Feature/missing and disabled named entities

ChiaraDipi:
Add configuration param `showLists` to deactivate lists button from main menu even if there named entities lists encoded in the source file. In this case the detail of entity will be shown only in the reference in the main text.

Handled situation when no named entities lists are found in the source file. In this case both the list button in main menu and detail of the reference in the main text are turned off.

### DIFF
--- a/src/app/components/named-entity-ref/named-entity-ref.component.ts
+++ b/src/app/components/named-entity-ref/named-entity-ref.component.ts
@@ -14,7 +14,9 @@ import { EVTModelService } from '../../services/evt-model.service';
 @register(NamedEntityRefData)
 export class NamedEntityRefComponent extends HighlightableComponent {
   @Input() data: NamedEntityRefData;
-
+  availableEntities$ = this.evtModelService.namedEntities$.pipe(
+    map(ne => ne.all.entities.length > 0),
+  );
   entity$ = this.evtModelService.namedEntities$.pipe(
     map(ne => ne.all.entities.find(e => e.id === this.data.entityId) || 'notFound'),
   );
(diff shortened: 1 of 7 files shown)


src/app/components/named-entity-ref/named-entity-ref.component.ts
```diff
@@ -14,7 +14,9 @@ import { EVTModelService } from '../../services/evt-model.service';
 @register(NamedEntityRefData)
 export class NamedEntityRefComponent extends HighlightableComponent {
   @Input() data: NamedEntityRefData;
-
+  entitiesAvailable$ = this.evtModelService.namedEntities$.pipe(
```
szenzaro, line 17: `availableEntities$`?

src/app/main-menu/main-menu.component.ts
```diff
@@ -30,12 +32,21 @@ export class MainMenuComponent implements OnInit, OnDestroy {
...
     public translate: TranslateService,
     private modalService: ModalService,
+    private evtModelService: EVTModelService,
   ) {
   }
 
   ngOnInit() {
     this.loadUiConfig();
     this.isOpened = true;
+    this.evtModelService.namedEntities$.subscribe(ne => {
```
szenzaro, line 42: Where do you unsubscribe?

src/app/main-menu/main-menu.component.ts
```diff
@@ -51,45 +62,44 @@ export class MainMenuComponent implements OnInit, OnDestroy {
...
-      {
-        id: 'projectInfo',
-        iconInfo: {
-          icon: 'info-circle',
-          additionalClasses: 'icon',
-        },
-        label: 'projectInfo',
-        // callback: () => this.openGlobalDialogInfo,
-        callback: () => this.openGlobalDialogInfo(),
+    this.dynamicItems.push({
```
szenzaro, line 65: why pushing strategy?

PR #65, opened 2020-06-07
Refactor/parse logic

szenzaro:
Refactoring of the parser logic


src/app/models/evt-models.ts
```diff
@@ -71,9 +79,9 @@ export class NamedEntity extends GenericElementData {
 
 export type NamedEntityLabel = string;
 
-export class NamedEntityInfo extends GenericElementData {
+export class NamedEntityInfo extends GenericElement {
     label: string;
-    content: Array<GenericElementData | NamedEntitiesList>;
+    // content: Array<GenericElementData | NamedEntitiesList>; // TODO: check if we can delete this line
```
ChiaraDipi, line 84: Since NamedEntitiesList is a GenericElement I think you can delete this line

src/app/services/evt-model.service.ts
```diff
@@ -80,15 +82,34 @@ export class EVTModelService {
     shareReplay(1),
   );
 
+  // WITNESSES
+  public readonly witnessesData$ = this.editionSource$.pipe(
+    map((source) => this.witnessesParser.parseWitnessesData(source)),
+    shareReplay(1),
+  );
+
+  public readonly witnesses$ = this.witnessesData$.pipe(
```
ChiaraDipi, line 91: Which is the difference between witnesses and witnessesData? I think we should understand it somehow from the name

src/app/services/xml-parsers/basic-parsers.ts
```diff
@@ -0,0 +1,131 @@
...
+
+export class ParagraphParser extends EmptyParser implements Parser<XMLElement> {
+    attributeParser = createParser(AttributeParser, this.genericParse);
+    parse(xml: XMLElement): Paragraph {
+        const attributes = this.attributeParser.parse(xml);
+        const paragraphComponent: Paragraph = {
+            type: Paragraph,
+            content: parseChildren(xml, this.genericParse),
+            attributes,
+            n: attributes.n ?? '-1', // TODO: Check it this can become a number
```
ChiaraDipi, line 48: Remove `'-1'`: if no `@n` attribute there should be no `n` value for paragraph. It can't become a number since in TEI its value is a `teidata.text` and can also accept a non-numeric label.

src/app/services/xml-parsers/basic-parsers.ts
```diff
@@ -0,0 +1,131 @@
...
+
+export class LBParser extends EmptyParser implements Parser<XMLElement> {
+    attributeParser = createParser(AttributeParser, this.genericParse);
+    parse(xml: XMLElement): Lb {
+        const attributes = this.attributeParser.parse(xml);
+        const { n, rend, facs } = attributes;
+
+        return {
+            id: getID(xml),
+            n: n || '', // TODO: check why paragraph defaults to -1 and lbs do not
```
ChiaraDipi, line 63: Remove `TODO`. The correct way to treat `@n` attribute it's this one. The one in paragraph parser is wrong.

src/app/services/evt-model.service.ts
giuliac89: I think we can delete witnessesData$

src/app/services/xml-parsers/named-entities-parser.service.ts
```diff
@@ -81,7 +80,7 @@ export class NamedEntitiesParserService {
               parsedList.content = parsedList.content.concat(parsedSubList.content);
               parsedList.relations = parsedList.relations.concat(parsedSubList.relations);
             } else {
-              parsedList.content.push(this.parseNamedEntity(child));
+              parsedList.content.push(parse(child) as NamedEntity);
```
giuliac89, line 83: Didn't we say to avoid "as" when we know the type?

src/app/services/xml-parsers/generic-parser.service.ts
```diff
@@ -1,171 +1,8 @@
...
-  private parseChildren(xml: XMLElement) {
-    return complexElements(xml.childNodes).map(child => this.parse(child as XMLElement));
-  }
-
-  public parseAttributes(xml: XMLElement): AttributesData {
-    return Array.from(xml.attributes)
-      .map(({ name, value }) => ({ [name === 'xml:id' ? 'id' : name.replace(':', '-')]: value }))
-      .reduce((x, y) => ({ ...x, ...y }), {});
-  }
+  parse(xml: XMLElement) { return parse(xml); }
```
ChiaraDipi, line 7: Why don't just delete `GenericParserService` since it only has a method that is simply a wrapper to `parse` function?

src/app/services/xml-parsers/structure-xml-parser.service.ts
```diff
@@ -61,11 +61,13 @@ export class StructureXmlParserService {
         console.warn('TODO: Decide how to handle text division when there are no <pb>s');
         const mainText = document.querySelector('text');
         const content = Array.from(mainText.childNodes);
-        const page: PageData = {
+        const page: Page = {
           id: `page_${new Date().getTime()}`,
           label: 'Main Text',
           originalContent: content as XMLElement[],
-          parsedContent: content.map(child => this.genericParserService.parse(child as XMLElement)).filter(c => !!c.content),
+          parsedContent: content.map(child => this.genericParserService.parse(child as XMLElement))
```
ChiaraDipi, line 68: Why don't just delete GenericParserService since it only has a method that is simply a wrapper to parse function? (Same for line 54 and for lines 66 and 75 of witnesses-parser.service.ts)

src/app/services/xml-parsers/generic-parser.service.ts
szenzaro: Cause we will add other functionalities, like loading state.

PR #69, opened 2020-06-17
Feature/character declaration

ChiaraDipi:


ChiaraDipi:
> Characters are not displayed in the UI. Maybe a problem with the font family?

Yes, it depends on the font used for the text. We will add Junicode font in a next PR

PR #75, opened 2020-07-04
Feature/custom logo

ChiaraDipi:


src/app/main-header/main-header.component.ts
```diff
@@ -27,6 +27,12 @@ export class MainHeaderComponent implements OnDestroy {
   public editionConfig: EditionConfig;
   get editionHome() { return normalizeUrl(this.editionConfig.editionHome); }
 
+  get logoUrl() {
+    const customLogo = AppConfig?.evtSettings?.files?.logoUrl ?? undefined;
+
+    return customLogo?.length > 0 ? customLogo : '/assets/images/logo.png';
```
szenzaro, line 33: `!!customLogo` would be better than length check

PR #83, opened 2020-09-02
Bugfix/panel grabbing from selector

ChiaraDipi:


giuliac89:
We can try to find a more usable solution. With this fix it is difficult to move the panels

PR #85, opened 2020-09-05
Feature/supplied

ChiaraDipi:


src/app/components/supplied/supplied.component.html
```diff
@@ -0,0 +1,4 @@
+<span *ngIf="editionLevel !== 'diplomatic'" class="{{data.class}}" [ngbPopover]="data?.reason"
```
szenzaro, line 1: Can this be generalized into a component useful also for surplus?

ChiaraDipi: What do you intend to generalize? The popover for the reason?

szenzaro: Let's keep it as it is, if more components other than surplus and supplied have the same visualisation, we can add an abstraction of this component

szenzaro (approved):

PR #88, opened 2020-09-18
Feature/add indexed db support

giacomocerre:


src/app/services/idb.service.ts
```diff
@@ -0,0 +1,34 @@
...
+    return this.db.add(data);
+  }
+
+  update(id, data) {
+    return this.db.update(id, data);
+  }
+
+  remove(id) {
+    return this.db.delete(id);
+  }
```
szenzaro, line 33: Why do we need to wrap db?
Do you expect the use of Dexie to change over time?

src/app/models/evt-models.ts
```diff
@@ -263,3 +263,7 @@ export class Choice extends GenericElement {
     originalContent: Array<ParseResult<GenericElement>>;
     normalizedContent: Array<ParseResult<GenericElement>>;
 }
+
+export interface AnnotationID  {
+    annotationID: number;
+}
```
szenzaro, line 269: You can put this inside idb service

src/app/app-idb.module.ts
```diff
@@ -0,0 +1,20 @@
...
+  providers: [
+    IdbService,
+  ],
+  bootstrap: [],
+})
+export class AppIdbCoreModule {
+  constructor(@Optional() @SkipSelf() parentModule: AppIdbCoreModule) {
+    if (parentModule) {
+      throw new Error(
+        'AppIdbCoreModule is already loaded. Import it in the AppModule only');
```
szenzaro, line 17: Fix parenthesis

src/app/app-idb.module.ts
```diff
@@ -0,0 +1,20 @@
...
+@NgModule({
+  declarations: [],
+  imports: [],
+  providers: [
+    IdbService,
+  ],
+  bootstrap: [],
+})
+export class AppIdbCoreModule {
+  constructor(@Optional() @SkipSelf() parentModule: AppIdbCoreModule) {
```
szenzaro, line 14: Why this? Check if this may have become obsolete.

szenzaro (changes requested):

src/app/models/evt-models.ts
giacomocerre: :+1:

src/app/app-idb.module.ts
giacomocerre: :+1:

src/app/services/idb.service.ts
giacomocerre: I remove them and I use the functions of the DB directly.

ChiaraDipi (approved):

szenzaro (approved):

PR #91, opened 2020-10-04
Feature/annotator selection text

giacomocerre:


ChiaraDipi (changes requested):
Can you generalize the `@HostListener('mouseup')` part in order to avoid code duplication? You could try using a directive

src/app/panels/sources-panel/sources-panel.component.html
```diff
@@ -1,4 +1,4 @@
-<evt-panel [secondary]="true">
+<evt-panel evtAnnotator [secondary]="true">
```
ChiaraDipi, line 1: I would put the `evtAnnotator` directive in the content div, so that the annotator works only on text and not on tools in the header/footer

src/app/panels/text-panel/text-panel.component.html
```diff
@@ -1,4 +1,4 @@
-<evt-panel [showSecondaryContent]="isSecondaryContentOpened()">
+<evt-panel evtAnnotator [showSecondaryContent]="isSecondaryContentOpened()">
```
ChiaraDipi, line 1: Same as previous comment

src/app/panels/version-panel/version-panel.component.html
```diff
@@ -1,4 +1,4 @@
-<evt-panel [comparable]="true" [secondary]="true" [closable]="true" (hide)="emitHide()">
+<evt-panel evtAnnotator [comparable]="true" [secondary]="true" [closable]="true" (hide)="emitHide()">
```
ChiaraDipi, line 1: Same as previous comment

src/app/panels/witness-panel/witness-panel.component.html
```diff
@@ -1,4 +1,4 @@
-<evt-panel [comparable]="true" [secondary]="true" [closable]="true" (hide)="emitHide()">
+<evt-panel evtAnnotator [comparable]="true" [secondary]="true" [closable]="true" (hide)="emitHide()">
```
ChiaraDipi, line 1: Same as previous comment

ChiaraDipi (changes requested):

src/app/services/annotator/annotator.service.ts
```diff
@@ -0,0 +1,14 @@
+import { Injectable } from '@angular/core';
+import { Subject } from 'rxjs';
+
+@Injectable({
+  providedIn: 'root',
+})
+export class AnnotatorService {
+  textSelection = new Subject();
```
szenzaro, line 8: Missing type for this subject. What is the type of `textSelection`?

szenzaro (changes requested):

src/app/services/annotator/annotator.service.ts
```diff
@@ -0,0 +1,15 @@
+import { Injectable } from '@angular/core';
+import { Subject } from 'rxjs';
+
+@Injectable({
+  providedIn: 'root',
+})
+export class AnnotatorService {
+  textSelection: Subject<object> = new Subject();
```
szenzaro, line 8: is the `textSelection` just an `object`?

src/app/services/annotator/annotator.service.ts
```diff
@@ -0,0 +1,15 @@
...
+
+  getTextSelection() {
+    const selection = window.getSelection();
```
szenzaro, line 11: no need of this variable declaration

src/app/services/annotator/annotator.service.ts
```diff
@@ -0,0 +1,15 @@
+import { Injectable } from '@angular/core';
+import { Subject } from 'rxjs';
+
+@Injectable({
+  providedIn: 'root',
+})
+export class AnnotatorService {
+  textSelection: Subject<object> = new Subject();
+
+  getTextSelection() {
```
szenzaro, line 10: the name suggests you return something, but you do not.
Either change the name or return te text selection

src/app/services/annotator/annotator.service.ts
```diff
@@ -0,0 +1,15 @@
...
+    const selection = window.getSelection();
+    this.textSelection.next(selection);
+    console.log(this.textSelection);
```
szenzaro, line 13: remove `console.log`

src/app/services/annotator/annotator.service.ts
szenzaro: it seems that this can be written as `textSelection = new Subject<Selection>();`

szenzaro (dismissed):

giuliac89 (changes requested):

giuliac89 (approved):

PR #92, opened 2020-10-04
Feature/critical text

giuliac89:


src/app/components/apparatus-entry/apparatus-entry.component.html
```diff
@@ -0,0 +1,3 @@
+<span class="evt-app-entry" [attr.data-variance]="variance$ | async">
+   <evt-content-viewer *ngIf="data.lemma" [content]="data.lemma" [editionLevel]="editionLevel"></evt-content-viewer>
```
ChiaraDipi, line 2: I would add the bind to `itemsToHighlight`, too.

src/app/components/apparatus-entry/apparatus-entry.component.ts
```diff
@@ -0,0 +1,27 @@
+import { Component, Input } from '@angular/core';
+import { map } from 'rxjs/operators';
+import { ApparatusEntry } from 'src/app/models/evt-models';
+import { register } from 'src/app/services/component-register.service';
+import { EVTModelService } from 'src/app/services/evt-model.service';
```
ChiaraDipi, line 5: l. 3-5 => Switch to relative paths

src/app/components/reading/reading.component.ts
```diff
@@ -0,0 +1,13 @@
+import { Component, Input } from '@angular/core';
+import { Reading } from 'src/app/models/evt-models';
+import { register } from 'src/app/services/component-register.service';
```
ChiaraDipi, line 3: l. 2-3 => Switch to relative paths

src/app/services/xml-parsers/app-parser.ts
```diff
@@ -28,6 +31,26 @@ export class RdgParser extends EmptyParser implements Parser<XMLElement> {
         return Array.from(rdg.childNodes)
             .map((child: XMLElement) => this.genericParse(child));
     }
+
+    private readingIsSignificant(rdg: XMLElement): boolean {
```
ChiaraDipi, line 35: Change to `isReadingSignificant`

src/app/services/xml-parsers/app-parser.ts
```diff
@@ -28,6 +31,26 @@ export class RdgParser extends EmptyParser implements Parser<XMLElement> {
...
+            }
+        }
+
+        return isSignificant;
+    }
+
+    private isSignificant(notSignificantReading: string[], attributes: NamedNodeMap): boolean {
+        return !Array.from(attributes).some(({ name, value }) => {
+            return notSignificantReading.includes(`${name}=${value}`);
+        });
```
ChiaraDipi, line 52: You can simplify into 
```
return !Array.from(attributes).some(({ name, value }) => notSignificantReading.includes(`${name}=${value}`));
```

ChiaraDipi (changes requested):

src/app/services/xml-parsers/apparatus-entries-parser.service.ts
```diff
@@ -15,4 +15,31 @@ export class ApparatusEntriesParserService {
...
+      signRdgsNumber[app] = signRdgs[app].length;
+    });
+
+    return signRdgsNumber;
+  }
+
+  public getAppVariance(signRdgsNum: { [key: string]: number }, witList: { [key: string]: Witness }) {
+    const appsVariance = {};
+    Object.keys(signRdgsNum).forEach((x) => {
+      appsVariance[x] = signRdgsNum[x] / Object.keys(witList).length;
```
szenzaro, line 40: Can `witList` have zero length?

giuliac89: I don’t think so, but I'm not 100% sure. I'll find out and let you know.

src/app/components/apparatus-entry/apparatus-entry.component.html
giuliac89: Isn’t it better to add it to the reading component?

src/app/services/xml-parsers/apparatus-entries-parser.service.ts
giuliac89: It's possible that there is no listWit in case it's a single-witness edition. In this case the variance value is useless.

szenzaro: Ok for the usefulness, but what will happen to this snippet of code in such case?

giuliac89: Returns an invalid value. I'm fixing it. :)

szenzaro (approved):

ChiaraDipi (approved):

PR #98, opened 2020-10-13
Feature/nav from entity to occurrence

ChiaraDipi:


giuliac89 (commented):
What about adding an anchor to the first occurrence? In this way the user doesn't have to scroll if the occurrence is not immediately visible

src/app/components/named-entity/named-entity-occurrence/named-entity-occurrence.component.ts
```diff
@@ -8,4 +12,20 @@ import { NamedEntityOccurrence } from '../../../models/evt-models';
...
+  @Input() entityId: string;
+
+  constructor(
+    private evtModelService: EVTModelService,
+    private evtStatusService: EVTStatusService,
+  ) {
+  }
+
+  goToOccurrenceRef(ref: NamedEntityOccurrenceRef) {
+    this.evtModelService.pages$.pipe(first()).subscribe(pages => {
```
szenzaro, line 24: If you are not interested in the error notification or in the predicate function, use `take(1)` instead of `first()`

szenzaro (commented):
Just a little comment on my part

giuliac89 (approved):

PR #101, opened 2020-10-31
Feature/page division

giuliac89:


src/app/services/xml-parsers/structure-xml-parser.service.ts
```diff
@@ -21,63 +21,121 @@ export class StructureXmlParserService {
...
-            // tslint:disable-next-line: no-string-literal
-            .filter(c => !!c['content']), // TODO: FIXME: fix property access
-        };
-        pages[page.id] = page;
-        pagesIndexes.push(page.id);
+        this.parseDocument(document, pages, pagesIndexes);
       }
-      console.log(pages);
+
+      console.log('### PAGES ###', pages);
```
szenzaro, line 32: Remove console logs

src/app/services/xml-parsers/structure-xml-parser.service.ts
```diff
@@ -21,63 +21,121 @@ export class StructureXmlParserService {
...
+      });
+
+      const page: Page = {
+        id: element.getAttribute('xml:id') || 'page_' + (pagesIndexes.length + 1),
+        label: element.getAttribute('n') || 'Page ' + (pagesIndexes.length + 1),
+        originalContent: pageContent,
+        parsedContent: this.parsePageContent(isPbFront, pageContent),
+      };
+      pages[page.id] = page;
+      pagesIndexes.push(page.id);
```
szenzaro, line 91: Why there is need of a page indexes if there is already `pages` with his keys?

src/app/services/xml-parsers/structure-xml-parser.service.ts
```diff
@@ -21,63 +21,121 @@ export class StructureXmlParserService {
...
+    });
+  }
+
+  parseDocument(document: XMLElement, pages: Map<Page>, pagesIndexes: string[]) {
+    // No <pb> used => TODO: Decide how to handle text division
+    console.warn('TODO: Decide how to handle text division when there are no <pb>s');
```
szenzaro, line 97: Use comments for this

src/app/services/xml-parsers/structure-xml-parser.service.ts
```diff
@@ -21,63 +21,121 @@ export class StructureXmlParserService {
...
+    const mainText = document.querySelector('text');
+    const content = Array.from(mainText.childNodes);
+    const page: Page = {
+      id: `page_${new Date().getTime()}`,
```
szenzaro, line 101: Is there a `getId` function elsewhere?
If yes use that, otherwise make one and use it

szenzaro (changes requested):

src/app/services/xml-parsers/structure-xml-parser.service.ts
ChiaraDipi: I think it's useful to retrieve pages exact order.

src/app/services/xml-parsers/structure-xml-parser.service.ts
ChiaraDipi: I wouldn't use the current time to define the id of pages without identification. In this way, every time you load the edition the id changes and you cannot load specific pages based on URL parameters. I would use the xpath, or just the position/count of the element

ChiaraDipi: The id will probably go on the URL. Is the current time the best way to define the id of "virtual" pages?

src/app/services/xml-parsers/structure-xml-parser.service.ts
```diff
@@ -21,63 +21,129 @@ export class StructureXmlParserService {
...
-      console.log(pages);
     }
 
     return {
       pages,
       pagesIndexes,
     };
   }
+
+  parseDocumentPages(pageElements: NodeListOf<Element>, pagesNumber: number, pages: Map<Page>, pagesIndexes: string[]) {
```
ChiaraDipi, line 39: I'm not sure that passing parsed pages and index maps as parameters is the best choice. I would actually expected to have a `parseDocumentPage` method that given an `XMLElement` returns a `Page` element, so 
that it could be used also for single XMLElements without affecting any parsed data collector variable. 
Something like:
```
parseDocumentPages(pageElement: XMLElement, totalPages: number, pageIndex: number) {
   return { ...PAGE...};
}
```
or
```
parseDocumentPages(pageElement: XMLElement, isFirsPage: boolean, isLastPage: boolean) {
   return { ...PAGE...};
}
```
In any case maybe if you prefer the current implementation, I would change the name of property `pagesNumber`, since it is not immediate that this is the total number of pages (or directly remove it since you can retrieve the total number of page elements directly from `pageElements`). And the name for `pages` and `pagesIndexes` too, to make it more explicit that they are the parsed data collectors.

src/app/services/xml-parsers/structure-xml-parser.service.ts
```diff
@@ -21,63 +21,129 @@ export class StructureXmlParserService {
...
+        label: element.getAttribute('n') || 'Page ' + (pagesIndexes.length + 1),
+        originalContent: pageContent,
+        parsedContent: this.parsePageContent(pageContent),
+      };
+      pages[page.id] = page;
+      pagesIndexes.push(page.id);
+    });
+  }
+
+  parseDocument(document: XMLElement, pages: Map<Page>, pagesIndexes: string[]) {
```
ChiaraDipi, line 77: Same as the previous comment, I would expected to have a `parseDocumentAsPage` method that given an XMLElement returns a `Page` element, so that it could be used also for single XMLElements without affecting any parsed data collector variable.

src/app/services/xml-parsers/structure-xml-parser.service.ts
```diff
@@ -21,63 +21,129 @@ export class StructureXmlParserService {
...
+
+    pages[page.id] = page;
+    pagesIndexes.push(page.id);
+  }
+
+  parsePageContent(pageContent: OriginalEncodingNodeType[]) {
+   const parsedContent = [];
+   pageContent.map((child: XMLElement) => {
+      if (isNestedInElem(child, 'front') || child.nodeName === 'front') {
+        if (child.nodeType === 3 || isNestedInElem(child, '', [{ key: 'type', value:'document_front' }])) {
```
ChiaraDipi, line 111: I would put this value in a separate property of the parser, so that it can eventually be changed if needed.

ChiaraDipi (changes requested):

src/app/services/xml-parsers/structure-xml-parser.service.ts
szenzaro: Shouldn't be the page number be part of `Page` definition? (or `orderKey`).
Otherwise, the order can be stored in in an ordered structure data instead of a `Map`.

szenzaro (changes requested):

src/app/services/xml-parsers/structure-xml-parser.service.ts
giuliac89: I think this is a better choice: 
`parseDocumentPage(pageElement: XMLElement, totalPages: number, pageIndex: number) {
   return { ...PAGE...};
}`
 but I need also the list of pageElements (or the next page) to retrieve the elements between two <pb> elements.

giuliac89: Ok for
```
parseDocumentPage(pageElement: XMLElement, totalPages: number, pageIndex: number) {
   return { ...PAGE...};
}
```
 but I need also the list of `pageElements` (or the next page) to retrieve the elements between two `<pb>`. I'm not sure how to handle this.

src/app/services/xml-parsers/structure-xml-parser.service.ts
giuliac89: Could the property name be `frontOriginalContentAttr` or what?

src/app/services/evt-model.service.ts
```diff
@@ -29,7 +29,7 @@ export class EVTModelService {
 
   public readonly pages$: Observable<Page[]> = this.editionSource$.pipe(
     map((source) => this.editionStructureParser.parsePages(source)),
-    map(editionStructure => editionStructure.pagesIndexes.map(pageId => editionStructure.pages[pageId])),
+    map(editionStructure => editionStructure.pages.map((p) => p)),
```
szenzaro, line 32: Remove map to identity

src/app/services/xml-parsers/structure-xml-parser.service.ts
```diff
@@ -14,70 +13,129 @@ export class StructureXmlParserService {
...
-      console.log(pages);
     }
 
     return {
       pages,
-      pagesIndexes,
     };
   }
+
+  parseDocumentPage(page: XMLElement, pageIndex: number, pagesCollection: NodeListOf<XMLElement>): Page {
```
szenzaro, line 39: maybe `page` parameter should be called `node` or `xmlNode` or something similar

src/app/services/xml-parsers/structure-xml-parser.service.ts
```diff
@@ -14,70 +13,129 @@ export class StructureXmlParserService {
...
+        pageContent = getElementsAfterTreeNode(page);
+      }
+
+      if (!isFirstPage && !isLastPage) {
+        pageContent = getElementsBetweenTreeNode(page, pagesCollection[pageIndex + 1]);
+      }
+    }
+
+    return {
+      id: page.getAttribute('xml:id') || 'page_' + pageIndex,
```
szenzaro, line 64: Do we have a general getID function?
if not a task about using a single `getID` function would be nice

src/app/services/xml-parsers/structure-xml-parser.service.ts
```diff
@@ -14,70 +13,129 @@ export class StructureXmlParserService {
...
+      label: page.getAttribute('n') || 'Page ' + pageIndex,
```
szenzaro, line 65: should `'Page'` or the whole label be translation ready?

src/app/services/xml-parsers/structure-xml-parser.service.ts
```diff
@@ -14,70 +13,129 @@ export class StructureXmlParserService {
...
+      originalContent: pageContent,
```
szenzaro, line 66: what is this used for?

src/app/services/xml-parsers/structure-xml-parser.service.ts
```diff
@@ -14,70 +13,129 @@ export class StructureXmlParserService {
...
-          id: `page_${new Date().getTime()}`,
-          label: 'Main Text',
-          originalContent: content as XMLElement[],
-          parsedContent: content.map(child => this.genericParserService.parse(child as XMLElement))
-            // tslint:disable-next-line: no-string-literal
-            .filter(c => !!c['content']), // TODO: FIXME: fix property access
-        };
-        pages[page.id] = page;
-        pagesIndexes.push(page.id);
+        this.parseDocument(document, pages);
```
szenzaro, line 30: I think we should always parse the whole document, why is this different from `parseDocumentPage`?

src/app/services/xml-parsers/structure-xml-parser.service.ts
```diff
@@ -14,70 +13,129 @@ export class StructureXmlParserService {
...
+        if (child.nodeType === 3 || isNestedInElem(child, '', [{ key: 'type', value: this.frontOriginalContentAttr }])) {
+          parsedContent.push(this.genericParserService.parse(child));
+         } else {
+          const frontOriginalContentChild = child.querySelectorAll(`[type=${this.frontOriginalContentAttr}]`);
+          if (child.querySelectorAll(`[type=${this.frontOriginalContentAttr}]`).length > 0) {
+            Array.from(frontOriginalContentChild).forEach((c) => parsedContent.push(this.genericParserService.parse(c as XMLElement)));
+          }
+          if (child.getAttribute('type') === this.frontOriginalContentAttr) {
+            parsedContent.push(this.genericParserService.parse(child));
+          }
```
szenzaro, line 112: This is a bit convoluted. Maybe comments or a refactor are needed

src/app/services/xml-parsers/structure-xml-parser.service.ts
```diff
@@ -14,70 +13,129 @@ export class StructureXmlParserService {
...
+        }
+      } else {
+        parsedContent.push(this.genericParserService.parse(child));
+      }
+    });
+
+   return parsedContent;
+  }
+
+  hasFrontOriginalContent(el: HTMLElement): boolean {
```
szenzaro, line 122: What is an "original content"?

szenzaro (changes requested):

PR #112, opened 2021-01-18
Feature/xi include support

ChiaraDipi:


src/app/services/edition-data.service.ts
```diff
@@ -30,11 +32,57 @@ export class EditionDataService {
...
+      xiIncludeLoadsSubs.push(
+        this.http.get(fileXInclude, { responseType: 'text' }).pipe(
+          tap((fileData) => {
+            let includedDoc: XMLElement;
+            if (typeof fileData !== 'object' && typeof fileData === 'string') {
+              includedDoc = parseXml(fileData);
+            } else {
+              includedDoc = fileData;
+            }
+            let includedTextElem;
```
szenzaro, line 56: missing type

src/app/services/edition-data.service.ts
```diff
@@ -30,11 +32,57 @@ export class EditionDataService {
...
+            if (fileXpointer) {
+              includedTextElem = doc.querySelector(`[*|id="${fileXpointer}"]`) || includedDoc.querySelector('text');
+            } else {
+              includedTextElem = includedDoc.querySelector('text');
+            }
+            element.parentNode.replaceChild(includedTextElem, element);
```
szenzaro, line 62: Why do you replace the element? 
Would it be viable to store the xi:included elements elsewhere (a map URL -> XML for example)?

src/app/services/edition-data.service.ts
```diff
@@ -30,11 +32,57 @@ export class EditionDataService {
 
         return editionDoc;
       }),
+      mergeMap((editionData) => this.loadXIinclude(editionData, editionUrl.substring(0, editionUrl.lastIndexOf('/') + 1))),
       publishReplay(1),
```
szenzaro, line 36: Why `publishReplay` instead of `shareReplay`?

src/app/services/edition-data.service.ts
```diff
@@ -30,11 +32,57 @@ export class EditionDataService {
...
       catchError(() => this.handleLoadingError()));
   }
 
+  loadXIinclude(doc: HTMLElement, baseUrlPath: string) {
+    const filesToInclude = Array.from(doc.getElementsByTagName('xi:include'));
+    const xiIncludeLoadsSubs = [];
+    filesToInclude.forEach(element => {
+      const fileXInclude = baseUrlPath + element.getAttribute('href');
+      const fileXpointer = element.getAttribute('xpointer');
+      xiIncludeLoadsSubs.push(
```
szenzaro, line 47: Try to avoid using `push`

szenzaro (changes requested):

src/app/services/edition-data.service.ts
```diff
@@ -30,11 +32,57 @@ export class EditionDataService {
...
+        this.http.get(fileXInclude, { responseType: 'text' }).pipe(
```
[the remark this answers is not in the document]
ChiaraDipi: It doesn't work without `{ responseType: 'text' }` and if I put both the explicit return type and `{ responseType: 'text' }` I have a TS error.

src/app/services/edition-data.service.ts
```diff
@@ -30,11 +32,57 @@ export class EditionDataService {
...
+          tap((fileData) => {
+            let includedDoc: XMLElement;
+            if (typeof fileData !== 'object' && typeof fileData === 'string') {
+              includedDoc = parseXml(fileData);
+            } else {
+              includedDoc = fileData;
```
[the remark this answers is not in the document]
ChiaraDipi: Just `HTMLElement` 😅. I'll fix it

src/app/services/edition-data.service.ts
ChiaraDipi: IMHO Once you've retrieved the content linked with xi:included you don't need that element anymore. It is used just to split very long documents in smaller and more manageable files, and document with xi:included elements should be considered as a single-file document.

src/app/services/edition-data.service.ts
ChiaraDipi: This was like this before, I've just added `mergeMap`. 
Anyway, `shareReplay` generates the following error which block the text from being shown.
Schermata 2021-01-19 alle 10 27 00

src/app/services/edition-data.service.ts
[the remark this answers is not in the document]
szenzaro: I was talking about the `get<T>` parameter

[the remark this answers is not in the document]
ChiaraDipi: I know. 
If I put `this.http.get<string>` I have an error with `{ responseType: 'text' }` and I have to remove it.
Schermata 2021-01-25 alle 09 28 27

But if I remove `{ responseType: 'text' }`, the http call does not work and I have this error:
Schermata 2021-01-25 alle 09 33 13

src/app/services/edition-data.service.ts
```diff
@@ -30,11 +32,49 @@ export class EditionDataService {
 
         return editionDoc;
       }),
+      mergeMap((editionData) => this.loadXIinclude(editionData, editionUrl.substring(0, editionUrl.lastIndexOf('/') + 1))),
```
szenzaro, line 35: Check edition doc type and fix `typeof source !== 'object' && typeof source === 'string'`

szenzaro (changes requested):

src/app/services/edition-data.service.ts
ChiaraDipi: If I use `shareReplay` then in the Network Tab **_n_** calls to xml files. They are all with code 304, so it should not be a problem, but I don't really like it, that's why originally was chosen `publishReplay(1)`, (in fact that is a part of code that was there before the current PR).

szenzaro (approved):

PR #113, opened 2021-01-26
Feature/critical text page division

giuliac89:


src/app/services/xml-parsers/structure-xml-parser.service.ts
```diff
@@ -24,7 +24,7 @@ export class StructureXmlParserService {
     const front: XMLElement = el.querySelector(this.frontTagName);
     const body: XMLElement = el.querySelector(this.bodyTagName);
 
-    const pbs = Array.from(el.querySelectorAll(this.pageTagName));
+    const pbs = Array.from(el.querySelectorAll(this.pageTagName)).filter((p) => !p.getAttribute('ed'));
```
szenzaro, line 27: Is this use of the attribute documented somewhere?

szenzaro (changes requested):

src/app/services/xml-parsers/structure-xml-parser.service.ts
giuliac89: Yes, the TEI guidelines referring to <pb> say that:

> The more specialized attributes break, ed, or edRef should be preferred when the intent is to indicate whether or not the page break is word-breaking, or to note the source from which it derives.

https://www.tei-c.org/release/doc/tei-p5-doc/en/html/ref-pb.html

src/app/view-modes/collation/collation.component.html
```diff
@@ -1,11 +1,7 @@
 <gridster [options]="options">
     <gridster-item [item]="textPanelItem">
-        <evt-text-panel
-            [pageID]="currentPageID$ | async" 
-            (pageChange)="changePage($event)"       
-            [hideEditionLevelSelector]="true" 
-            [editionLevelID]="'critical'"
-        ></evt-text-panel>
+        <evt-critical-text-panel
+            [editionLevelID]="'critical'"></evt-critical-text-panel>
```
ChiaraDipi, line 4: Since you use a specific component that is `evt-critical-text-panel` you probably don't need to pass the editionLevelID as input. For this specific panel I expect the edition level will never change and always be "critical". Can't you define it directly in the component?

src/app/view-modes/reading-text/reading-text.component.html
```diff
@@ -1,11 +1,15 @@
...
+        <evt-text-panel *ngIf="(currentEditionLevel$ | async) !== 'critical'"
+            [pageID]="currentPageID$ | async"
             [editionLevelID]="currentEditionLevel$ | async"
-            (pageChange)="changePage($event)" 
+            [hidePageSelector]="(currentEditionLevel$ | async) === 'critical'"
+            (pageChange)="changePage($event)"
             (editionLevelChange)="changeEditionLevel($event)"
         ></evt-text-panel>
+        <evt-critical-text-panel *ngIf="(currentEditionLevel$ | async) === 'critical'"
+            [editionLevelID]="currentEditionLevel$ | async"
```
ChiaraDipi, line 11: Same as before

src/app/view-modes/text-sources/text-sources.component.html
```diff
@@ -1,11 +1,15 @@
...
-            [pageID]="currentPageID$ | async" 
+        <evt-text-panel *ngIf="(currentEditionLevel$ | async) !== 'critical'"
+            [pageID]="currentPageID$ | async"
             [editionLevelID]="currentEditionLevel$ | async"
+            [hidePageSelector]="(currentEditionLevel$ | async) === 'critical'"
             (pageChange)="changePage($event)" 
             (editionLevelChange)="changeEditionLevel($event)"
         ></evt-text-panel>
+        <evt-critical-text-panel *ngIf="(currentEditionLevel$ | async) === 'critical'"
+            [editionLevelID]="currentEditionLevel$ | async"
```
ChiaraDipi, line 11: Same as before

src/app/view-modes/text-text/text-text.component.html
```diff
@@ -11,6 +12,7 @@
         <evt-text-panel 
             [pageID]="currentPageID$ | async" 
             [editionLevelID]="currentEditionLevels[1]"
+            [hidePageSelector]="currentEditionLevels[1] === 'critical'"
             (pageChange)="changePage($event)"     
             (editionLevelChange)="changeEditionLevel($event, 2)"
         ></evt-text-panel>
```
ChiaraDipi, line 18: Why in this case if it is critical edition level you don't use the `evt-critical-text-panel`?

src/app/view-modes/text-versions/text-versions.component.html
```diff
@@ -1,11 +1,15 @@
...
-            [pageID]="currentPageID$ | async" 
+        <evt-text-panel *ngIf="(currentEditionLevel$ | async) !== 'critical'"
+            [pageID]="currentPageID$ | async"
             [editionLevelID]="currentEditionLevel$ | async"
+            [hidePageSelector]="(currentEditionLevel$ | async) === 'critical'"
             (pageChange)="changePage($event)" 
             (editionLevelChange)="changeEditionLevel($event)"
         ></evt-text-panel>
+        <evt-critical-text-panel *ngIf="(currentEditionLevel$ | async) === 'critical'"
+            [editionLevelID]="currentEditionLevel$ | async"
```
ChiaraDipi, line 11: Same as before

ChiaraDipi (changes requested):
Actually I'm not so sure that the solution of creating a panel dedicated to critical text is the ideal solution. It is simply a textual panel that changes only in the part of the text where you don't have a single page, but more pages one after the other. The other elements (toolbar, footer, secondary panels) are more or less always the same. Moreover, you don't have to distinguish between `<evt-text-panel>` and `<evt-critical-text-panel>` in every view that should show the text.

src/app/view-modes/text-text/text-text.component.html
giuliac89: It's a mistake. In critical edition the text-text view mode is disabled. Do you agree or could it be useful?

giuliac89:
@ChiaraDipi I've tried to manage the critical text in `<evt-text-panel>`, but the behavior is different in some view modes. In the text-image view mode the critical text must follow normal pages division. Maybe I can retrieve the view mode from the text panel component and manage this cases here. What do you think about it?

ChiaraDipi:
> @ChiaraDipi I've tried to manage the critical text in `<evt-text-panel>`, but the behavior is different in some view modes. In the text-image view mode the critical text must follow normal pages division. Maybe I can retrieve the view mode from the text panel component and manage this cases here. What do you think about it?

Yeah, right. I forgot about this possibility.
Thus it's ok as you did.
In any case, I you not have an `@Input` parameter for `editionLevelID` for `evt-criticalt-text-panel` since by the name I expect it to always show the critical edition level.

src/app/view-modes/text-text/text-text.component.html
ChiaraDipi: I don't know. You can leave it like as it is and we will change it in the future if we encounter such a problem.

src/app/services/xml-parsers/structure-xml-parser.service.ts
szenzaro: The description do not refer to "filtering out" pbs that have `ed` attribute though.
I am not sure this is the right thing to do

src/app/view-modes/text-versions/text-versions.component.html
```diff
@@ -1,9 +1,10 @@
 <gridster [options]="options">
     <gridster-item [item]="textPanelItem">
-        <evt-text-panel
-            [pageID]="currentPageID$ | async" 
-            [editionLevelID]="currentEditionLevel$ | async"
-            (pageChange)="changePage($event)" 
+        <evt-text-panel *ngIf="currentEditionLevel$ | async as currentEditionLevel"
+            [pageID]="currentPageID$ | async"
+            [editionLevelID]="currentEditionLevel"
+            [hidePageSelector]="currentEditionLevel === 'critical'"
```
ChiaraDipi, line 6: See previous comment about `hidePageSelector` property.

src/app/view-modes/reading-text/reading-text.component.html
```diff
@@ -1,9 +1,10 @@
 <gridster [options]="layoutOptions">
     <gridster-item [item]="textPanelItem">
-        <evt-text-panel 
-            [pageID]="currentPageID$ | async" 
-            [editionLevelID]="currentEditionLevel$ | async"
-            (pageChange)="changePage($event)" 
+        <evt-text-panel *ngIf="currentEditionLevel$ | async as currentEditionLevel"
+            [pageID]="currentPageID$ | async"
+            [editionLevelID]="currentEditionLevel"
+            [hidePageSelector]="currentEditionLevel === 'critical'"
```
ChiaraDipi, line 6: See previous comment about `hidePageSelector` property.

ChiaraDipi (changes requested):

src/app/panels/text-panel/text-panel.component.ts
```diff
@@ -14,6 +15,7 @@ import { EvtIconInfo } from '../../ui-components/icon/icon.component';
 })
 export class TextPanelComponent implements OnInit, OnDestroy {
   @Input() hideEditionLevelSelector: boolean;
+  @Input() hidePageSelector: boolean;
```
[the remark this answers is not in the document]
giuliac89: Ok, it makes sense, but I think is not right to hide the page selector in the case of absence of pages, because there is the possibility that there is some original content in front matter (case in which we create a further dummy page), so I would keep it in all cases.

szenzaro (approved):

PR #120, opened 2021-02-05
Feature/parser register

szenzaro:
Change parser definition workflow


giuliac89 (changes requested):
It seems to be fine, just two small comments to check

giuliac89 (approved):

PR #121, opened 2021-02-08
Feature/encoding desc parser

ChiaraDipi:


src/app/models/evt-models.ts
```diff
@@ -849,3 +849,129 @@ export class SourceDesc extends GenericElement {
...
+    unitDecl: Array<ParseResult<GenericElement>>; // TODO: Add specific type when unitDecl is handled
+    schemaSpec: Array<ParseResult<GenericElement>>; // TODO: Add specific type when schemaSpec is handled
+    schemaRef: Array<ParseResult<GenericElement>>; // TODO: Add specific type when schemaRef is handled
+}
+
+export class ProjectDesc extends GenericElement {
+    content: Paragraph[];
+}
+
+export class SamplingDecl extends GenericElement {
```
szenzaro, line 872: Use type alias for this

src/app/models/evt-models.ts
```diff
@@ -849,3 +849,129 @@ export class SourceDesc extends GenericElement {
...
+    content: Paragraph[];
+}
+
+export type CorrectionStatus = 'high' | 'medium' | 'low' | 'unknown';
+export type CorrectionMethod = 'silent' | 'markup';
+export class Correction extends GenericElement {
```
szenzaro, line 878: Extend `ProjectDesc`?

src/app/models/evt-models.ts
```diff
@@ -849,3 +849,129 @@ export class SourceDesc extends GenericElement {
...
+    content: Paragraph[];
+    status?: CorrectionStatus;
+    method?: CorrectionMethod;
+}
+
+export type NormalizationMethod = 'silent' | 'markup';
+export class Normalization extends GenericElement {
```
szenzaro, line 885: Extend `ProjectDesc`?

src/app/models/evt-models.ts
```diff
@@ -849,3 +849,129 @@ export class SourceDesc extends GenericElement {
...
+
+export class Segmentation extends GenericElement {
+    content: Paragraph[];
+}
+
+export class StdVals extends GenericElement {
+    content: Paragraph[];
+}
+
+export class Interpretation extends GenericElement {
```
szenzaro, line 919: Type alias these

szenzaro (changes requested):
Missing Changelog update

src/app/models/evt-models.ts
ChiaraDipi: If I use type alias (`export type SamplingDecl = ProjectDesc;`), then I get an error in the parser in the setting of `type` property:
```
parse(xml: XMLElement): SamplingDecl {
    return {
      ...super.parse(xml),
      type: SamplingDecl,
      content: queryAndParseElements<Paragraph>(xml, 'p'),
    };
  }
```
The error says: «'SamplingDecl' only refers to a type, but is being used as a value here.»

src/app/models/evt-models.ts
ChiaraDipi: See previous comment

szenzaro (approved):

PR #126, opened 2021-02-18
Feature/ms identifier component

saramaenza:


ChiaraDipi (changes requested):
I would try to add a new component in ui-components in order to generalize the HTML part of label + content.
In this way you can move the general css rules from styles.scss to the scss of the component.
I imagine this component (es `headerSection` or `projectInfoSection`) to have at least an input for the label and the class name and to use the transclusion with `ng-content` for the content of the section. You can take the generic PanelComponent component and the more specific TextPanelComponent as an example.

With this component you can then simplify the HTML of the `MsIdentifierComponent` template you will have for example:
```
<evt-header-section *ngIf="data.idno.length>0" [label]="'idno'" [sectionClass]="'idno'">
   <div content>   
      <evt-content-viewer *ngFor="let i of data.idno" [content]="i"></evt-content-viewer>
   </div>
</evt-header-section>
```
and you will have a unique point where to change the structure HTML or css classes if needed.

src/app/components/ms-identifier/ms-identifier.component.html
```diff
@@ -0,0 +1,71 @@
+<div class="msIdentifier">
+    <p class="projectInfo-sectionSubHeader">{{ 'msidentifier' | translate }}</p>
+    <div>
+        <evt-project-info-section *ngIf="data?.settlement.length>0 || data?.country.length>0 || data?.region.length>0" [label]="'place'" [sectionClass]="'settlement'">
```
szenzaro, line 4: The condition of `*ngIf` seems redundant: the inner components re-evaluate the same conditions.
Check if this OR condition is needed.

src/app/components/ms-identifier/ms-identifier.component.html
```diff
@@ -0,0 +1,71 @@
...
+            <span content>   
+                <evt-content-viewer *ngFor="let s of data.settlement" [content]="s"></evt-content-viewer>
```
szenzaro, line 6: can data be undefined since you  checked with `data?.xxx` before?
The same comment applies for every `data.xxx` occurrence.

src/app/components/ms-identifier/ms-identifier.component.html
```diff
@@ -0,0 +1,71 @@
...
+            <span content>   
+               <evt-content-viewer *ngFor="let c of data.collection" [content]="c" class="projectInfo-block"></evt-content-viewer>
+            </span>
+        </evt-project-info-section>
+        <evt-project-info-section *ngFor="let m of data?.msName | keyvalue" [label]="'msName'" [sectionClass]="'msName'">
+            <span content>   
+               <evt-content-viewer [content]="data.msName[m.key]"></evt-content-viewer>
+            </span>
+        </evt-project-info-section>
+        <evt-project-info-section *ngIf="data?.idno.length>0" [label]="'idno'" [sectionClass]="'idno'">
```
szenzaro, line 37: Can `idno` be undefined?

src/app/components/ms-identifier/ms-identifier.component.html
ChiaraDipi: I think the `*ngIf` condition is necessary because if settlement, country and region are all empty we should not even print the label "place".

src/app/components/ms-identifier/ms-identifier.component.html
```diff
@@ -0,0 +1,74 @@
...
+        </evt-ms-desc-section>
+        <evt-ms-desc-section *ngIf="data?.collection.length>0" [label]="'collection'" [additionalClass]="'collection'" [inlineLabel]="'inlineLabel'">
+            <div content>   
+               <evt-content-viewer *ngFor="let c of data.collection" [content]="c" class="projectInfo-block"></evt-content-viewer>
+            </div>
+        </evt-ms-desc-section>
+        <evt-ms-desc-section *ngFor="let m of data?.msName | keyvalue" [label]="'msName'" [additionalClass]="'msName'" [inlineLabel]="'inlineLabel'">
+            <div content>   
+               <evt-content-viewer [content]="data.msName[m.key]"></evt-content-viewer>
+            </div>
```
ChiaraDipi, line 39: You don't need the keyvalue pipe; just do `*ngFor="let m of data?.msName"` and then bind `content` directly to `m`.

src/app/components/ms-identifier/ms-identifier.component.html
```diff
@@ -0,0 +1,74 @@
...
+        </evt-ms-desc-section>
+        <evt-ms-desc-section *ngIf="data?.idno.length>0" [label]="'idno'" [additionalClass]="'idno'" [inlineLabel]="'inlineLabel'">
+            <div content>   
+               <evt-content-viewer *ngFor="let i of data.idno" [content]="i" class="projectInfo-block"></evt-content-viewer>
+            </div>
+        </evt-ms-desc-section>
+        <span *ngFor="let a of data?.altIdentifier | keyvalue" >
+            <evt-ms-desc-section [label]="'altIdentifier'" [inlineLabel]="'inlineLabel'">
+                <div content>   
+                    <evt-content-viewer></evt-content-viewer>
```
ChiaraDipi, line 49: This `evt-content-viewer` is not binded to any content data. Check it.

src/app/components/ms-identifier/ms-identifier.component.html
```diff
@@ -0,0 +1,74 @@
...
+            <div content>   
+               <evt-content-viewer [content]="data.msName[m.key]"></evt-content-viewer>
+            </div>
+        </evt-ms-desc-section>
+        <evt-ms-desc-section *ngIf="data?.idno.length>0" [label]="'idno'" [additionalClass]="'idno'" [inlineLabel]="'inlineLabel'">
+            <div content>   
+               <evt-content-viewer *ngFor="let i of data.idno" [content]="i" class="projectInfo-block"></evt-content-viewer>
+            </div>
+        </evt-ms-desc-section>
+        <span *ngFor="let a of data?.altIdentifier | keyvalue" >
```
ChiaraDipi, line 46: As before, you don't need the keyvalue pipe; just do `*ngFor="let a of data?.altIdentifier"` and then use directly `a` for bindings instead of `data.altIdentifier[a.key]`

src/app/components/ms-identifier/ms-identifier.component.html
```diff
@@ -0,0 +1,74 @@
+<div class="msIdentifier">
+    <evt-ms-desc-section [label]="'msIdentifier'">
+        <div content>   
+            <evt-content-viewer></evt-content-viewer>
+        </div>
+    </evt-ms-desc-section>
+    <div>
+        <evt-ms-desc-section [label]="'place'" [additionalClass]="'settlement'" [inlineLabel]="'inlineLabel'">
```
ChiaraDipi, line 8: `inlineLabel` is a boolean input. You should pass `true` or `false` to it, not a string! (Check every occurence of it)

src/styles.scss
```diff
@@ -178,3 +178,21 @@ body {
...
+    margin-top: 14px;
+}
+
+.projectInfo-blockLabel {
+    display: block;
+    margin-bottom: 2px;
+    font-variant: small-caps;
+    color: #666;
+    font-size: 90%;
+}
```
ChiaraDipi, line 198: You probably don't need those two rules any more, check it

ChiaraDipi (changes requested):

src/app/components/ms-identifier/ms-identifier.component.html
```diff
@@ -0,0 +1,74 @@
+<div class="msIdentifier">
```
szenzaro, line 1: Refactor `Identifier` section as a component

szenzaro (changes requested):

szenzaro (approved):

src/app/components/identifier/identifier.component.html
```diff
@@ -0,0 +1,24 @@
+<div [class.has-list-class]="listClass">
+    <evt-ms-desc-section [label]="'place'" [additionalClass]="'settlement'" [inlineLabel]="true">
```
giuliac89, line 2: The label is displayed even if it has no value.
Example in https://www.tei-c.org/release/doc/tei-p5-doc/en/html/examples-altIdentifier.html

src/app/ui-components/ms-desc-section/ms-desc-section.component.scss
```diff
@@ -0,0 +1,34 @@
...
+        font-variant: small-caps;
+        color: #666;
+    }
+    &.has-inline-label {
+        display: flex;
+        .section-label {
+            margin-right: 5px;
+        }
+        .section-content {
+            display: inline;
```
giuliac89, line 14: This rule is useless. You don't need it using flexbox.

src/app/ui-components/ms-desc-section/ms-desc-section.component.scss
```diff
@@ -0,0 +1,34 @@
+.section {
+    display: block;
```
giuliac89, line 2: I think this rule is useless, check it.

src/app/ui-components/ms-desc-section/ms-desc-section.component.scss
```diff
@@ -0,0 +1,34 @@
...
+    .section-label {
+        display: inline;
```
giuliac89, line 4: This rule is useless. You don't need it using flexbox.

src/app/components/ms-identifier/ms-identifier.component.html
```diff
@@ -0,0 +1,23 @@
...
+        <evt-identifier [data]="data"></evt-identifier>
+        <evt-ms-desc-section *ngIf="data?.institution" [label]="'institution'" [additionalClass]="'institution'" [inlineLabel]="true">  
+            <evt-content-viewer [content]="data.institution" content></evt-content-viewer>
+        </evt-ms-desc-section>
+        <evt-ms-desc-section *ngIf="data?.msName.length>0" [label]="'msName'" [additionalClass]="'msName'" [inlineLabel]="true"> 
+            <evt-content-viewer *ngFor="let m of data.msName" [content]="m" class="projectInfo-block" content></evt-content-viewer>
+        </evt-ms-desc-section>
+        <span *ngFor="let a of data?.altIdentifier">
+            <evt-ms-desc-section [label]="'altIdentifier'" [inlineLabel]="true"> 
+                <evt-content-viewer [content]="data?.altIdentifier" content></evt-content-viewer>
```
giuliac89, line 15: In which case is it necessary? What value can the altIdentifier label have?

giuliac89 (changes requested):

szenzaro (approved):

giuliac89 (approved):

PR #138, opened 2021-04-20
Feature/additional component

saramaenza:


src/app/models/evt-models.ts
```diff
@@ -637,7 +637,7 @@ export class CustodialHist extends GenericElement {
     structuredData: boolean;
     pEl?: Paragraph[];
     ab?: Array<ParseResult<GenericElement>>; // TODO: Add specific type when ab is handled
-    custEvent?: CustEvent;
+    custEvent?: CustEvent[];
```
szenzaro, line 640: Is it possible to remove the optional field and assume that when not provided this field is a zero length array?

src/app/models/evt-models.ts
```diff
@@ -637,7 +637,7 @@ export class CustodialHist extends GenericElement {
     structuredData: boolean;
     pEl?: Paragraph[];
```
szenzaro, line 638: Is it possible to remove the optional field and assume that when not provided this field is a zero length array?

src/app/models/evt-models.ts
```diff
@@ -637,7 +637,7 @@ export class CustodialHist extends GenericElement {
...
     ab?: Array<ParseResult<GenericElement>>; // TODO: Add specific type when ab is handled
```
szenzaro, line 639: Is it possible to remove the optional field and assume that when not provided this field is a zero length array?

src/app/components/additional/additional.component.html
```diff
@@ -0,0 +1,59 @@
+<div class="additional">
+    <evt-ms-desc-section [label]="'additional'"></evt-ms-desc-section>
+    <ng-container *ngIf="data?.adminInfo.structuredData; else elseBlock">
```
szenzaro, line 3: Avoid using `else`

src/app/components/additional/additional.component.html
```diff
@@ -0,0 +1,59 @@
...
+        <div class="flex-container" *ngIf="data?.adminInfo?.recordHist">
+            <div class="flex-item-label">
```
szenzaro, line 5: Can the class property be used directly on the next line? (evt-ms-desc-section)

szenzaro (changes requested):
Missing Changelog update

src/app/models/evt-models.ts
giuliac89: I agree, but I think this optimization can be handled in another pr.

src/app/models/evt-models.ts
giuliac89: I agree, but I think this optimization can be handled in another pr.

src/app/models/evt-models.ts
giuliac89: I agree, but I think this optimization can be handled in another pr.

src/app/components/additional/additional.component.html
```diff
@@ -0,0 +1,59 @@
...
+                <evt-ms-desc-section [label]="'recordHist'" [inlineLabel]="true"></evt-ms-desc-section>
+            </div>
+            <div class="flex-item-information">
+                <evt-content-viewer [content]="data.adminInfo.recordHist.source" class="projectInfo-block"></evt-content-viewer> 
+                <evt-content-viewer *ngFor="let c of data.adminInfo.recordHist.change" [content]="c"></evt-content-viewer>
```
giuliac89, line 10: `<change>` element is an unhandled element. Currently the visualization of unhandled elements is not optimized, because they are all displayed inline and there are no spaces between them.  I think we can change it in the following way (example for this case):
- Remove `flat()` from `unhandledElement()` function in parser-models.ts
- Change `change?: Array<ParseResult<GenericElement>>;` in `changes?: Array<ParseResult<GenericElement>>;`
- In template will be 
```
<div *ngFor="let changes of data.adminInfo.recordHist.change">
      <evt-content-viewer *ngFor="let change of changes" [content]="change"></evt-content-viewer>
</div>
```

In this way the visualization of unhandled elements wil be better and you won't need too many changes when adding the parser.

@szenzaro do you think it can work?
This optimization can be handled in another pr.

src/app/components/additional/additional.component.scss
```diff
@@ -0,0 +1,36 @@
...
+.flex-item-information {
+    flex-basis: 100%;
+    margin-bottom: 0.7rem;
+}
+
+::ng-deep p {
+    margin-bottom: 0;
+}
+
+@media (min-width: 1400px) {
```
giuliac89, line 28: The behavior of the style is different from the other components. Check it

giuliac89 (changes requested):

szenzaro (approved):

giuliac89 (approved):

PR #153, opened 2021-06-10
Feature/msdesc component

saramaenza:


src/app/models/evt-models.ts
```diff
@@ -348,7 +348,7 @@ export class MsPart extends MsFrag {
 
 export class MsDesc extends MsPart {
     id: string;
-    msFrag: MsFrag;
+    msFrag: MsFrag[];
```
szenzaro, line 351: Maybe `msFrags` since it is an array?

szenzaro (changes requested):
Missing CHANGELOG update

szenzaro (approved):

PR #155, opened 2021-06-13
Feature/msdesc header button

saramaenza:


src/app/panels/text-panel/text-panel.component.ts
```diff
@@ -66,6 +70,15 @@ export class TextPanelComponent implements OnInit, OnDestroy {
     shareReplay(1),
   );
 
+  public msDesc$ = this.evtModelService.editionSource$.pipe(
+    map(s => {
+      const msDesc = s.querySelector<HTMLElement>('msDesc');
+      const msDescParser = createParser(MsDescParser, parse);
+
+      return msDescParser.parse(msDesc);
```
szenzaro, line 78: This should be handled by a parser. Why is there parser logic inside a component?

src/app/panels/text-panel/text-panel.component.ts
```diff
@@ -90,12 +103,24 @@ export class TextPanelComponent implements OnInit, OnDestroy {
...
+      this.secondaryContent = '';
+    }
+    else {
+      this.showSecondaryContent = false;
+      this.msDescOpen = false;
+    }
+  }
   getSecondaryContent(): string {
     return this.secondaryContent;
   }
```
szenzaro, line 126: No need of this, just access the class field.

src/app/panels/text-panel/text-panel.component.html
```diff
@@ -25,7 +30,10 @@
                 [itemsToHighlight]="itemsToHighlight$ | async"></evt-page>
         </ng-container>
     </div>
-    <div secondary-content>{{ getSecondaryContent() }}</div>
+    <div secondary-content>
+        <evt-ms-desc *ngIf="msDescOpen" [data]="msDesc$ | async"></evt-ms-desc>
+        {{ getSecondaryContent() }}
```
szenzaro, line 35: Just access the class field

szenzaro (changes requested):
Missing CHANGELOG changes

src/app/panels/text-panel/text-panel.component.ts
giuliac89: Parser logic should never be in a component. Move it to evt-model-service.

giuliac89 (changes requested):

giuliac89 (approved):

PR #164, opened 2021-12-04
Feature/image image component

saramaenza:


src/app/components/osd/osd.component.ts
```diff
@@ -111,13 +113,16 @@ export class OsdComponent implements AfterViewInit, OnDestroy {
...
-      this._page = v;
+      this._page = v + 1;
       this.pageChange.next(this._page);
     }
   }
   get page() { return this._page; }
   @Output() pageChange = new EventEmitter<number>();
 
+  // tslint:disable-next-line: no-any
+  @Output() data: EventEmitter<any> = new EventEmitter<any>();
```
szenzaro, line 124: Type is inferred implicitly. No need to specify it again.

Is this emitter of type any?

What "data" stands for, this variable name is probably too generic?

src/app/components/osd/osd.component.ts
```diff
@@ -111,13 +113,16 @@ export class OsdComponent implements AfterViewInit, OnDestroy {
   private _page: number;
   @Input() set page(v: number) {
     if (v !== this._page) {
-      this._page = v;
+      this._page = v + 1;
```
szenzaro, line 116: Why "+1"? The page should not be changed from the param initial value.

src/app/components/osd/osd.component.ts
```diff
@@ -185,11 +190,14 @@ export class OsdComponent implements AfterViewInit, OnDestroy {
 
         this.viewer.addHandler('page', ({ page }) => {
           this.pageChange.next(page + 1);
+          this._page = page;
+          this.data.emit(page);
```
szenzaro, line 194: We already have a page change event. When and why do you need a new one?

src/app/panels/image-panel/image-panel.component.html
```diff
@@ -1,14 +1,34 @@
...
+    </ng-select>
+    <evt-ms-desc-selector (selectionChange)="setMsDescID($event)" (msDescOpen)="isMsDescOpen($event)"></evt-ms-desc-selector>
     </div>
     <div content>
         <evt-osd
             *ngIf="manifest"
-            [manifestURL]="manifest">
+            [manifestURL]="manifest"
+            [page]="imageIndex"
+            (data)="eventHandler($event)"
```
szenzaro, line 22: Since you are vinding to page change event use that instead.

src/app/panels/image-panel/image-panel.component.ts
```diff
@@ -1,13 +1,93 @@
...
+    // tslint:disable-next-line: no-any
+    this.http.get(`${this.manifest}`).subscribe((item: any) => {
+      item.sequences.forEach((value) => {
+        this.images.push(...value.canvases);
+      });
+      this._imageIndex = 0;
+      this.currentImg = this.images[this.page].label;
+      this.cdref.detectChanges();
+
+      return this.images;
```
szenzaro, line 57: The returned value is never used, do you need it?

src/app/panels/image-panel/image-panel.component.ts
```diff
@@ -1,13 +1,93 @@
...
+  public page = 0;
+
+  toggleImg(event){
+    this._imageIndex = this.images.indexOf(event);
+  }
+
+  eventHandler(event: number){
+    this.page = event;
+    this.currentImg = this.images[this.page].label;
+    this.cdref.detectChanges();
```
szenzaro, line 44: Why do you call explicitly change detection?

src/app/panels/image-panel/image-panel.component.ts
```diff
@@ -1,13 +1,93 @@
...
+
+  ngOnInit() {
+    // tslint:disable-next-line: no-any
+    this.http.get(`${this.manifest}`).subscribe((item: any) => {
+      item.sequences.forEach((value) => {
+        this.images.push(...value.canvases);
+      });
+      this._imageIndex = 0;
+      this.currentImg = this.images[this.page].label;
+      this.cdref.detectChanges();
```
szenzaro, line 55: Why do you call explicitly change detection?

src/app/panels/image-panel/image-panel.component.ts
```diff
@@ -1,13 +1,93 @@
...
+      this.secondaryContent = newContent;
+      this.msDescOpen = false;
+    }
+    else {
+      this.showSecondaryContent = false;
+      this.secondaryContent = '';
+    }
+  }
+
+  isMsDescOpen(event: boolean){
```
szenzaro, line 81: This should be just `this.msDescOpen` why do you need the other lines?

szenzaro: Weird to have a check on the internal state that depends on an event that is a boolean.
Check this function again, it is probably overcomplicated.

src/app/panels/image-panel/image-panel.component.ts
```diff
@@ -1,13 +1,93 @@
...
+    this.showSecondaryContent = event;
+    if (this.showSecondaryContent){
+      this.msDescOpen = true;
+      this.secondaryContent = '';
+    }
+  }
+
+  setMsDescID(event: string){
```
szenzaro, line 89: Check naming, of parameters, you call `event` boolean, strings and actual events

szenzaro: this could be a setter

src/app/view-modes/image-image/image-image.component.ts
```diff
@@ -0,0 +1,35 @@
...
+  public layoutOptions: GridsterConfig = {};
+  public imagePanel1Item: GridsterItem = { cols: 1, rows: 1, y: 0, x: 0 };
+  public imagePanel2Item: GridsterItem = { cols: 1, rows: 1, y: 0, x: 1 };
+
+  ngOnInit() {
+    this.initGridster();
+  }
+
+  private initGridster() {
+    this.layoutOptions = {
```
szenzaro, line 19: you can put this at line 10.

src/app/view-modes/image-image/image-image.component.ts
```diff
@@ -0,0 +1,35 @@
...
+  templateUrl: './image-image.component.html',
+  styleUrls: ['./image-image.component.scss'],
+})
+export class ImageImageComponent implements OnInit {
+  public layoutOptions: GridsterConfig = {};
+  public imagePanel1Item: GridsterItem = { cols: 1, rows: 1, y: 0, x: 0 };
+  public imagePanel2Item: GridsterItem = { cols: 1, rows: 1, y: 0, x: 1 };
+
+  ngOnInit() {
+    this.initGridster();
```
szenzaro, line 15: Do you need to set layout options here or is it enough to set at construction time?

szenzaro (changes requested):

src/app/panels/image-panel/image-panel.component.ts
saramaenza: I added change detection because otherwise the error "ERROR Error: NG0100: ExpressionChangedAfterItHasBeenCheckedError: Expression has changed after it was checking" would appear in the console, since every time the image name is updated without the page being refreshed

src/app/panels/image-panel/image-panel.component.ts
saramaenza: I added change detection because otherwise the error "ERROR Error: NG0100: ExpressionChangedAfterItHasBeenCheckedError: Expression has changed after it was checking" would appear in the console, since every time the image name is updated without the page being refreshed

src/app/components/osd/osd.component.ts
saramaenza: this.data.emit (page); is used to keep track of page changes through the arrow buttons

src/app/panels/image-panel/image-panel.component.html
saramaenza: page change event does not consider the page change through arrows

src/app/components/osd/osd.component.ts
szenzaro: it is unified. The change page event should be agnostic with regards to the reason it has been changed

src/app/panels/image-panel/image-panel.component.html
szenzaro: make the buttons fire that event instead

src/app/panels/image-panel/image-panel.component.ts
szenzaro: Usually such errors is due to an improper initialisation.

src/app/components/osd/osd.component.ts
```diff
@@ -112,12 +114,14 @@ export class OsdComponent implements AfterViewInit, OnDestroy {
   @Input() set page(v: number) {
     if (v !== this._page) {
       this._page = v;
-      this.pageChange.next(this._page);
+      this.pageChange.next(this._page + 1);
```
szenzaro, line 117: Why the `+1`?

src/app/components/osd/osd.component.ts
```diff
@@ -185,11 +189,14 @@ export class OsdComponent implements AfterViewInit, OnDestroy {
 
         this.viewer.addHandler('page', ({ page }) => {
           this.pageChange.next(page + 1);
+          this._page = page;
+          this.pageImg.emit(page);
```
szenzaro, line 193: Here you could just use the `pageChange` event

src/app/components/osd/osd.component.ts
```diff
@@ -185,11 +189,14 @@ export class OsdComponent implements AfterViewInit, OnDestroy {
 
         this.viewer.addHandler('page', ({ page }) => {
           this.pageChange.next(page + 1);
+          this._page = page;
```
szenzaro, line 192: You should never change `_page` explicitly` it will create different values between pageChange and the actual page. You should use the event emitter IMHO

src/app/panels/image-panel/image-panel.component.ts
```diff
@@ -1,13 +1,72 @@
...
+    });
+  }
+
+  isSecondaryContentOpened(): boolean {
+    return this.showSecondaryContent;
+  }
+
+  isMsDescOpen(event: boolean){
+    this.showSecondaryContent = event;
+    this.msDescOpen = event;
```
szenzaro, line 65: If they are the same, why do you need two properties?

src/app/panels/image-panel/image-panel.component.html
```diff
@@ -1,14 +1,34 @@
...
+    </ng-select>
+    <evt-ms-desc-selector (selectionChange)="setMsDescID($event)" (msDescOpen)="isMsDescOpen($event)"></evt-ms-desc-selector>
     </div>
     <div content>
         <evt-osd
             *ngIf="manifest"
-            [manifestURL]="manifest">
+            [manifestURL]="manifest"
+            [page]="imageIndex"
+            (pageImg)="eventPageImg($event)"
```
szenzaro, line 22: Just use `(page)=...`

szenzaro (changes requested):

src/app/components/osd/osd.component.ts
saramaenza: By simply using the `pageChange` event, it is not possible to select the image with index 0 (the front cover) because the page counter starts at 1.

src/app/components/osd/osd.component.ts
ChiaraDipi: @saramaenza can answer to @szenzaro? Why do you need the `+1`?

src/app/components/osd/osd.component.ts
```diff
@@ -174,4 +174,5 @@ export class OsdComponent implements AfterViewInit, OnDestroy {
   ngOnDestroy(): void {
     this.subscriptions.forEach((s) => s.unsubscribe());
   }
+
```
ChiaraDipi, line 177: Remove this line

src/app/panels/image-panel/image-panel.component.html
```diff
@@ -1,21 +1,33 @@
-<evt-panel [showSecondaryContent]="msDescOpen"
-    [hideHeader]="false"
-    [hideFooter]="true">
+<evt-panel [showSecondaryContent]="isSecondaryContentOpened()">
     <div header-left>
-        <!-- TODO: Add dropdowns for navigation -->
-        <evt-ms-desc-selector #msDesc (selectionChange)="setMsDescID($event)" (msDescOpen)="setMsDescOpen($event)"></evt-ms-desc-selector>
+    <ng-select
```
ChiaraDipi, line 3: Can you restore the TODO comment?

src/app/panels/image-panel/image-panel.component.html
```diff
@@ -1,21 +1,33 @@
...
+            (pageChange)="eventPageImg($event)">
         </evt-osd>
         <p *ngIf="!viewerData">Found no source file</p>
     </div>
-
     <div secondary-content>
-        <div *ngIf="msDescOpen">
-            <evt-ms-desc [data]="currentMsDesc$ | async"></evt-ms-desc>    
+        <div *ngIf="(msDesc$ | async) as msDescs">
+            <div *ngFor="let msDesc of msDescs">
```
ChiaraDipi, line 28: You can use the async binding directly in the loop:
```
<div *ngFor="let msDesc of msDescs$ | async">
```

FYI, if you need to add a level to handle logic (like you did here), but don't need additional HTML you can use `ng-container`.

src/app/panels/image-panel/image-panel.component.ts
```diff
@@ -12,24 +11,74 @@ import { EVTModelService } from '../../services/evt-model.service';
...
+  public selectedPage;
+  public msDescID = '';
+  public value: number;
+  public images = [];
+  public currentImg: string;
+  // tslint:disable-next-line: variable-name
+  private _imageIndex: number;
+  public manifest;
+  public imagePath;
+  public filename;
```
ChiaraDipi, line 33: Move all these property before the contructor.

src/app/panels/image-panel/image-panel.component.ts
ChiaraDipi: You can try to use Observable for currentImage and let the async method handle it.

src/app/panels/image-panel/image-panel.component.ts
```diff
@@ -12,24 +11,74 @@ import { EVTModelService } from '../../services/evt-model.service';
...
+
+  toggleImg(event){
+    this._imageIndex = this.images.indexOf(event);
+  }
+
+  eventPageImg(pageImg: number){
```
ChiaraDipi, line 39: The name of this method is not very explanatory. Can you find another one?

src/app/panels/image-panel/image-panel.component.ts
```diff
@@ -12,24 +11,74 @@ import { EVTModelService } from '../../services/evt-model.service';
...
+  public value: number;
+  public images = [];
+  public currentImg: string;
+  // tslint:disable-next-line: variable-name
+  private _imageIndex: number;
+  public manifest;
+  public imagePath;
+  public filename;
+
+  toggleImg(event){
```
ChiaraDipi, line 35: Add space before "{"

src/app/panels/image-panel/image-panel.component.ts
```diff
@@ -12,24 +11,74 @@ import { EVTModelService } from '../../services/evt-model.service';
...
+    if (this.viewerData?.type === 'default' || this.viewerData?.type === 'graphics'){
+      this.images = this.viewerData?.value?.xmlImages;
+      this.imagePath =  this.viewerData?.value?.xmlImages[0]?.url;
+      this.filename = this.imagePath.split('/').pop().split('.')[0];
+      this.currentImg = this.filename;
+    }
+    else {
+      this.currentImg = 'Error loading';
+    }
   }
```
ChiaraDipi, line 70: All the logic here depends on `viewerData`. You can try to change the `@Input` prop into a setter/getter and move this logic inside it.
```
private _vd: ViewerDataType;
@Input() set viewerData(vd: ViewerDataType) {
   // Your logic goes here
}
get viewerData() {
   return this._vd;
}
```
NB: The getter is only needed if you access `viewerData` elsewhere, otherwise you can omit it.

src/app/panels/image-panel/image-panel.component.ts
```diff
@@ -12,24 +11,74 @@ import { EVTModelService } from '../../services/evt-model.service';
...
 
-  setMsDescID(msDescId: string) {
-    this.currentMsDescId$.next(msDescId);
+  isSecondaryContentOpened(): boolean {
+    return this.showSecondaryContent;
   }
+
+  isMsDescOpen(event: boolean){
```
ChiaraDipi, line 76: The name of this method suggests a return value. If you don't need a return value probably you should change the name of the method.

src/app/view-modes/image-image/image-image.component.ts
ChiaraDipi: You can probably move it in the constructor!

src/app/view-modes/image-image/image-image.component.ts
ChiaraDipi: If you move it at line 15 (15 is the new 10 XD), you can remove the method `initGridster`

ChiaraDipi (changes requested):
Since you added a new icon, you should update also the file `evt-icons-project.json` into `src/assets/fonts`, otherwise new future icons addition will override yours!

PR #168, opened 2021-12-12
Bugfix/move ms desc sel in img panel

ChiaraDipi:
In previous versions of EVT, manuscript info were available only in image panel. Thus the msDesc selector had to be removed from text panel and put in image panel.


src/app/panels/image-panel/image-panel.component.html
```diff
@@ -1,8 +1,9 @@
-<evt-panel
-    [hideHeader]="true"
+<evt-panel [showSecondaryContent]="msDescOpen"
+    [hideHeader]="false"
     [hideFooter]="true">
     <div header-left>
         <!-- TODO: Add dropdowns for navigation -->
```
giuliac89, line 5: Is the comment still necessary?

giuliac89 (changes requested):

src/app/panels/image-panel/image-panel.component.html
ChiaraDipi: Yes. We need to remember to add dropdown for page navigation

giuliac89 (changes requested):
This pr can be merged after resolving the conflict

szenzaro (approved):

src/app/components/ms-desc-selector/ms-desc-selector.component.html
```diff
@@ -7,5 +7,5 @@
     [items]="msDesc$ | async" #ngSelectComponent
     (clear)="resetMsDesc();"
     [placeholder]="'selectMsDesc' | translate" 
-    (change)="msDescID = $event.id; openMsDescContent()">     
+    (change)="msDescID = $event?.id; openMsDescContent()">
```
szenzaro, line 10: Is it possible to get an undefined event on change event occurrence?

szenzaro (approved):

src/app/components/ms-desc-selector/ms-desc-selector.component.html
ChiaraDipi: Yes, you can unselect everything (in this case the box containing msDesc will be close). The change event of ng-select will output the whole model, thus if you unselect everything, the model will be undefined.

giuliac89 (approved):

PR #171, opened 2021-12-24
Feature/image viewer from xml

giacomocerre:


ChiaraDipi (changes requested):
I think there's a case you haven't handled: in previous version of EVT we used to support also editions without surface informations. In this case we relied on page ids to build the path to the image, along with the path to images folder properly configured and the file extension (we indicated in Manual that in this case images should have all the same extension and be put all in a single folder).

src/app/view-modes/image-text/image-text.component.ts
```diff
@@ -26,9 +33,28 @@ export class ImageTextComponent implements OnInit {
...
+  getImageViewerType(editionImages: EditionImages, surface: Surface[]): ViewerDataType {
+    for (const key in editionImages) {
+      if (editionImages[key].enabled && editionImages[key] !== '' && !!editionImages[key]){
+        if (key === 'manifest'){
+          return {type: key, value: {url: editionImages[key].value}};
+        }
+        const xmlImages = [].concat.apply([], surface.map((s) =>
+          s[editionImages[key].value]
+            ? ({url: `${AppConfig.evtSettings.files.imagesFolderUrl}/${s.corresp}.jpg`})
+            : s[key].map((a) => ({widht: a.attributes.width, height: a.attributes.height, url: `assets/data/${a.attributes.url}`})),
```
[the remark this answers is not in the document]
giacomocerre: I have not used the path in conf because in the URL attribute there may already be the complete path.

src/app/components/osd/osd.component.ts
```diff
@@ -96,16 +117,25 @@ export class OsdComponent implements AfterViewInit, OnDestroy {
...
-  // tslint:disable-next-line: variable-name
-  private _manifestURL: string;
-  @Input() set manifestURL(v: string) {
-    if (v !== this._manifestURL) {
-      this._manifestURL = v;
-      this.manifestURLChange.next(this._manifestURL);
+  private _viewerDataType: string; // tslint:disable-line: variable-name
+  private _manifestPath: string; // tslint:disable-line: variable-name
+  private _xmlImagesPath: XMLImagesValues[]; // tslint:disable-line: variable-name
+  @Input() set data(v: ViewerDataType) {
```
szenzaro, line 123: `data` is a really generic name

src/app/components/osd/osd.component.ts
```diff
@@ -96,16 +117,25 @@ export class OsdComponent implements AfterViewInit, OnDestroy {
...
     }
+    if (v.value.xmlImages !== this._xmlImagesPath) {
+      this._xmlImagesPath = v.value.xmlImages;
+      this.xmlImageChange.next(this._xmlImagesPath);
+    }
+    this._viewerDataType = v.type;
   }
-  get manifestURL() { return this._manifestURL; }
-  manifestURLChange = new BehaviorSubject(undefined);
+  get viewerDataType() { return this._viewerDataType; }
```
szenzaro, line 134: is there a reason to not make `_viewerDataType` public?

src/app/components/osd/osd.component.ts
```diff
@@ -96,16 +117,25 @@ export class OsdComponent implements AfterViewInit, OnDestroy {
...
+  get manifestURL() { return this._manifestPath; }
+  get xmlImageURL() { return this._xmlImagesPath; }
```
szenzaro, line 136: Same as `viewerDAtaType`, if you want to access these members why do you keep them private?

src/app/components/osd/osd.component.ts
```diff
@@ -150,10 +170,35 @@ export class OsdComponent implements AfterViewInit, OnDestroy {
...
+
   ngAfterViewInit() {
     this.viewerId = uuid('openseadragon');
     this.div.nativeElement.id = this.viewerId;
 
+    if (this.viewerDataType === 'manifest'){
+      this.tileSources = this.manifestTileSources;
+    }else{
+      this.tileSources = this.xmlImageTileSources;
+    }
```
szenzaro, line 200: Better use polymorphic objects

src/app/panels/image-panel/image-panel.component.ts
```diff
@@ -1,13 +1,14 @@
...
   styleUrls: ['./image-panel.component.scss'],
 })
-export class ImagePanelComponent {
-  manifest = AppConfig.evtSettings.files.manifestURL !== '' && !!AppConfig.evtSettings.files.manifestURL
-    ? AppConfig.evtSettings.files.manifestURL
-    : undefined;
+export class ImagePanelComponent{
+
+  // tslint:disable-next-line: variable-name
+  @Input() data: ViewerDataType;
```
szenzaro, line 12: `data` is not a real good name

szenzaro (changes requested):

src/app/models/evt-polymorphic-models.ts
```diff
@@ -0,0 +1,132 @@
...
+            width: manifestResource.width,
+        };
+    }
+}
+
+class XMLSource {
+
+    getDataType(key: string, data: Surface[]): ViewerDataType {
+        const XMLImages: XMLImagesValues[] = [];
+        data.map(s =>
```
ChiaraDipi, line 93: Since you're not changing the original array, you can use `forEach`. 
Otherwise you can do
```
XMLImages = data.map(s =>
            s[AppConfig.evtSettings.files.editionImages[key].value]
                ? {
                    url: `${AppConfig.evtSettings.files.imagesFolderUrl}/${s.corresp}.jpg`
                } : {
                    width: s[key][0].width,
                    height: s[key][0].height,
                    url: `${AppConfig.evtSettings.files.imagesFolderUrl}/${s[key][0].url}`
                });
```

src/app/models/evt-polymorphic-models.ts
```diff
@@ -0,0 +1,132 @@
...
+            s[AppConfig.evtSettings.files.editionImages[key].value]
+              ? XMLImages.push(({ url: `${AppConfig.evtSettings.files.imagesFolderUrl}/${s.corresp}.jpg` }))
+              : XMLImages.push({ width: s[key][0].width, height: s[key][0].height, url: `${AppConfig.evtSettings.files.imagesFolderUrl}/${s[key][0].url}` }));
+
+        return { type: key, value: { xmlImages: XMLImages } };
```
ChiaraDipi, line 98: If you rename the variable into `xmlImages`, you can simplify this into
```
return {
   type: key,
   value: { xmlImages },
};
```

src/app/models/evt-polymorphic-models.ts
```diff
@@ -0,0 +1,132 @@
...
+    getTileSource<K extends keyof T>(change: BehaviorSubject<ViewerDataInput>, type: string | K, http?: HttpClient): ReturnType<T[K]['getTileSource']> {
+        return this.factories[type].getTileSource(change, http) as ReturnType<T[K]['getTileSource']>;
+    }
+    getDataType<K extends keyof T>(type: string, data?: Surface[]): ReturnType<T[K]['getDataType']> {
+        return this.factories[type].getDataType(type, data) as ReturnType<T[K]['getDataType']>;
+    }
+}
+
+class ManifestSource {
+
```
ChiaraDipi, line 52: Remove empty line

src/app/models/evt-polymorphic-models.ts
```diff
@@ -0,0 +1,132 @@
...
+            profile: [manifestResource.service['@profile']],
+            protocol: 'http://iiif.io/api/image',
+            height: manifestResource.height,
+            width: manifestResource.width,
+        };
+    }
+}
+
+class XMLSource {
+
```
ChiaraDipi, line 90: Remove empty line

src/app/panels/image-panel/image-panel.component.ts
```diff
@@ -1,13 +1,14 @@
...
   selector: 'evt-image-panel',
   templateUrl: './image-panel.component.html',
   styleUrls: ['./image-panel.component.scss'],
 })
-export class ImagePanelComponent {
-  manifest = AppConfig.evtSettings.files.manifestURL !== '' && !!AppConfig.evtSettings.files.manifestURL
-    ? AppConfig.evtSettings.files.manifestURL
-    : undefined;
+export class ImagePanelComponent{
+
```
ChiaraDipi, line 10: Remove empty line

src/app/panels/image-panel/image-panel.component.ts
```diff
@@ -1,13 +1,14 @@
...
+  // tslint:disable-next-line: variable-name
```
ChiaraDipi, line 11: Do you still need this comment?

src/app/view-modes/image-text/image-text.component.ts
```diff
@@ -1,10 +1,14 @@
 import { Component, OnInit } from '@angular/core';
 import { DisplayGrid, GridsterConfig, GridsterItem, GridType } from 'angular-gridster2';
 import { map, shareReplay } from 'rxjs/operators';
-import { Page } from 'src/app/models/evt-models';
-import { EVTStatusService } from 'src/app/services/evt-status.service';
-import { EditionLevel } from '../../app.config';
+import { ViewerSource } from 'src/app/models/evt-polymorphic-models';
```
ChiaraDipi, line 4: Use relative path

src/app/view-modes/image-text/image-text.component.ts
```diff
@@ -1,10 +1,14 @@
...
+import { AppConfig, EditionImages, EditionLevel } from '../../app.config';
+import { Page, Surface, ViewerDataType, XMLImagesValues } from '../../models/evt-models';
+import { register } from '../../services/component-register.service';
+import { EVTModelService } from '../../services/evt-model.service';
+import { EVTStatusService } from '../../services/evt-status.service';
 
+@register(Surface)
```
ChiaraDipi, line 11: Why do you connect the `Surface` to this component? `ImageTextComponent` is actually a view mode that is connected to a specific route.

src/app/view-modes/image-text/image-text.component.ts
```diff
@@ -26,9 +34,25 @@ export class ImageTextComponent implements OnInit {
 
   constructor(
     private evtStatusService: EVTStatusService,
+    private evtModelService: EVTModelService,
   ) {
   }
 
+  getImageViewerType(editionImages: EditionImages, surface: Surface[]): ViewerDataType {
+    for (const key of Object.keys(editionImages)) {
```
ChiaraDipi, line 42: change into `Object.keys(editionImages).forEach`

src/app/models/evt-polymorphic-models.ts
```diff
@@ -0,0 +1,132 @@
...
+    getDataType(key: string, data: Surface[]): ViewerDataType {
+        const XMLImages: XMLImagesValues[] = [];
+        data.map(s =>
+            s[AppConfig.evtSettings.files.editionImages[key].value]
+              ? XMLImages.push(({ url: `${AppConfig.evtSettings.files.imagesFolderUrl}/${s.corresp}.jpg` }))
+              : XMLImages.push({ width: s[key][0].width, height: s[key][0].height, url: `${AppConfig.evtSettings.files.imagesFolderUrl}/${s[key][0].url}` }));
```
ChiaraDipi, line 96: I think you should check url: if it points to an online resource you should not add `imagesFolderUrl` before it.

src/assets/config/file_config.json
```diff
@@ -1,5 +1,19 @@
...
+		"graphics": {
+			"value": "url",
+			"enabled": true
+		},
+		"surface": {
+			"value": "corresp",
+			"enabled": true
+		},
+		"default" : true
+	},
```
ChiaraDipi, line 17: As default configuration we should enable only one of this data source

ChiaraDipi (changes requested):

ChiaraDipi:
If "default" method is used, the image doesn't change when the page changes

ChiaraDipi (changes requested):

src/app/utils/js-utils.ts
```diff
@@ -96,3 +96,9 @@ export function isBoolString(s: string) {
 export function snakeToCamelCased(str) {
     return str.replace(/-([a-z])/g, (g) => g[1].toUpperCase());
 }
+
+export function isUrl(path: string, location?: string) {
+    const regexp = /(ftp|http|https):\/\/(\w+:{0,1}\w*@)?(\S+)(:[0-9]+)?(\/|\/([\w#!:.?+=&%@!\-\/]))?/;
+
+    return regexp.test(path) ? path : `${location ? location + '/' : '' }${path}`;
```
ChiaraDipi, line 103: I don't like it like this:
- Why if the regex test doesn't pass it is ok to pass location+path? Who tells me that the string thus formed is correct?
- since the function name is a question, I expect to have a boolean as returning value.
I would leave just one param (`path`) and return the result of the test with the regex.
```
export function isUrl(path: string, location?: string) {
   return /(ftp|http|https):\/\/(\w+:{0,1}\w*@)?(\S+)(:[0-9]+)?(\/|\/([\w#!:.?+=&%@!\-\/]))?/.test(path);
}
```

Of course, you will have to write more code where you use it
```
s[AppConfig.evtSettings.files.editionImagesSource[key].value]
    ? {
        url: isUrl(s.corresp) ? s.corresp : localImagesFolder + s.corresp,
    } : {
        width: s[key][0].width,
        height: s[key][0].height,
        url: isUrl(s[key][0].url) ? s[key][0].url : localImagesFolder + s[key][0].url,
    });
```
but I think it is better that utils functions are as more generic as possible.

ChiaraDipi (changes requested):

szenzaro (approved):

ChiaraDipi (approved):

PR #172, opened 2021-12-29
Feature/apparatus entries box

giuliac89:


src/app/components/apparatus-entry/apparatus-entry-detail/apparatus-entry-detail.component.ts
```diff
@@ -0,0 +1,57 @@
...
+  get significantRdg(): Reading[] {
+    return this.data.readings.filter((rdg) => rdg.significant);
+  }
+
+  get notSignificantRdg(): Reading[] {
+    return this.data.readings.filter((rdg) => !rdg.significant);
+  }
+
+  get readings(): Reading[] {
+    return [this.data.lemma].concat(this.significantRdg).concat(this.notSignificantRdg);
```
szenzaro, line 41: `return [this.data.lemma, ...this.significantRdg, ...this.notSignificantRdg]`

src/app/components/apparatus-entry/apparatus-entry-detail/apparatus-entry-detail.component.ts
```diff
@@ -0,0 +1,57 @@
...
+import { EVTModelService } from '../../../services/evt-model.service';
+@Component({
+  selector: 'evt-apparatus-entry-detail',
+  templateUrl: './apparatus-entry-detail.component.html',
+  styleUrls: ['./apparatus-entry-detail.component.scss'],
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+
+@register(ApparatusEntryDetailComponent)
+export class ApparatusEntryDetailComponent {
```
szenzaro, line 13: this seems to contain duplicated code from the app/rdg parsers. If so, try to refactor this.

src/app/components/apparatus-entry/apparatus-entry-detail/apparatus-entry-detail.component.html
```diff
@@ -0,0 +1,104 @@
+<div class="card app-detail-container">
```
szenzaro, line 1: This component is very large. Is it possible to modularize it a bit?

src/app/components/apparatus-entry/apparatus-entry.component.ts
```diff
@@ -1,26 +1,59 @@
...
+  public nestedApps: ApparatusEntry[] = [];
+
   variance$ = this.evtModelService.appVariance$.pipe(
     map((variances) => variances[this.data.id]),
+    shareReplay(1),
   );
 
   constructor(
     private evtModelService: EVTModelService,
+    @Optional() @SkipSelf() private parentDetailComponent?: ApparatusEntryDetailComponent,
```
szenzaro, line 28: Are these decorators needed?

src/app/components/apparatus-entry/apparatus-entry.component.ts
```diff
@@ -1,26 +1,59 @@
...
   ) {
-    super();
+    this.isInsideAppDetail = this.parentDetailComponent ? true : false;
+  }
+
+  ngOnInit() {
+    if (this.data.hasNestedApp) {
+      this.getNestedApps(this.data);
```
szenzaro, line 35: The name `getNestedApps` suggests that some value is returned. Instead it is a procedure.

src/app/components/apparatus-entry/apparatus-entry.component.ts
```diff
@@ -1,26 +1,59 @@
...
+    });
+  }
+
+  toggleAppEntryBox(e: MouseEvent) {
+    e.stopPropagation();
+    this.opened = !this.opened;
+  }
+
+  closeAppEntryBox() {
+    if (this.opened) {
```
szenzaro, line 55: you can just set it to false since if this.opened is true it will change to false, otherwise it will remain false

src/app/models/evt-models.ts
```diff
@@ -159,12 +158,15 @@ export interface WitnessGroup {
...
     groupId: string;
 }
+
 export class ApparatusEntry extends GenericElement {
     id: string;
     lemma: Reading;
     readings: Reading[];
     notes: Note[];
     originalEncoding: string;
+    hasNestedApp: boolean;
```
szenzaro, line 168: Isn't this equivalent to `nestedAppsIDs.length`?

szenzaro (changes requested):
Missing ChangeLog

ChiaraDipi (changes requested):
I agree with szenzaro comments and I would also try to modularize the ApparatusEntryDetail component if possible

src/app/components/apparatus-entry/apparatus-entry.component.ts
giuliac89: @Optional() it's necessary because normally an `apparatusEntry` doesn't have a parent which is an `ApparatusEntryDetailComponent`. This only applies to nested apps.
@SkipSelf seems useless (I can't understand why I added it)

src/app/components/apparatus-entry/apparatus-entry-detail/apparatus-entry-detail.component.html
ChiaraDipi: I would manage modularization in a new distinct feature.

src/app/components/apparatus-entry/apparatus-entry-detail/apparatus-entry-detail.component.ts
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  get rdgMetadata() {
+    return Object.keys(this.data.attributes).filter((key) => key !== 'id')
+      .reduce((obj, key) => {
+        obj[key] = this.data.attributes[key];
+
+        return obj;
+      },      {});
+  }
```
ChiaraDipi, line 37: I think it can be changed into 
```
return Object.keys(this.data.attributes).filter((key) => key !== 'id')
  .reduce(
    (obj, key) => ({
      ...obj,
      [key]: this.data.attributes[key],
    }),
    {});
```

src/app/components/apparatus-entry/apparatus-entry-detail/apparatus-entry-detail.component.ts
```diff
@@ -0,0 +1,73 @@
...
+
+  getNestedAppLemma(appId: string): Reading {
+    return this.nestedApps.find((c) => c.id === appId).lemma;
+  }
+
+  getNestedAppPos(appId: string): number {
+    const currentApp = this.nestedApps.find(nesApp => nesApp.id === appId);
+
+    return this.nestedApps.indexOf(currentApp);
```
ChiaraDipi, line 71: Try using directly 
```
this.nestedApps.findIndex(nesApp => nesApp.id === appId);
```

src/app/components/apparatus-entry/apparatus-entry-readings/apparatus-entry-readings.component.ts
```diff
@@ -0,0 +1,43 @@
...
+  get significantRdg(): Reading[] {
+    return this.data.readings.filter((rdg) => rdg.significant);
+  }
+
+  getWits$(witID: string): Observable<string[]> {
+    return this.groups$.pipe(
+      map((groups) => {
+        const currentGroup = groups.filter((g) => g.id === witID);
+
+        return currentGroup.map((g) => g.witnesses).reduce((x, y) => ([ ...x, ...y ]), []);
```
ChiaraDipi, line 38: Since it is an array you can name the variable "plural": `currentGroups`. Otherwise you can chain everything in one single line:
```
map((groups) =>
  groups.filter((g) => g.id === witID).map((g) => g.witnesses).reduce((x, y) => ([...x, ...y]), [])
),
```

src/app/services/xml-parsers/app-parser.ts
```diff
@@ -52,8 +52,7 @@ export class RdgParser extends EmptyParser implements Parser<XMLElement> {
         return !Array.from(attributes).some(({ name, value }) => notSignificantReading.includes(`${name}=${value}`));
     }
 }
-
-@xmlParser('app', AppParser)
+@xmlParser('evt-apparatus-entry-parser', AppParser)
```
ChiaraDipi, line 55: Why this change in tagName definition?

ChiaraDipi (changes requested):

src/app/services/xml-parsers/app-parser.ts
giuliac89: @ChiaraDipi I think I changed it to use the new `ParserRegister.get` function. See commit 6a6494d

ChiaraDipi (approved):

PR #180, opened 2022-04-09
Feature/link text to image

giacomocerre:


src/app/panels/image-panel/image-panel.component.ts
```diff
@@ -22,9 +36,16 @@ export class ImagePanelComponent {
 
   constructor(
     private evtModelService: EVTModelService,
+    private evtStatus: EVTStatusService,
   ) {
   }
 
+  updatePage(viewerPage: number){
+    this.evtModelService.pages$.pipe(take(1)).subscribe(
```
szenzaro, line 44: Create a new subscription at every page change creates a memory leak.
Ensure there is a single subscription (e.g. inside the constructor) and chain the event stream of the osd-component to an observable property of this component.

szenzaro (changes requested):

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Updated to Angular 9
 
 ### Added
+- Link between text and images
```
ChiaraDipi, line 13: This message is a little bit misleading: it seems that you have implemented the Image Text Linking feature. 
Change into something like "Text/images connection in page change"

src/app/panels/image-panel/image-panel.component.ts
```diff
@@ -1,17 +1,31 @@
 import { Component, Input } from '@angular/core';
 import { BehaviorSubject, combineLatest } from 'rxjs';
-import { filter, map } from 'rxjs/operators';
-import { ViewerDataType } from '../../models/evt-models';
+import { filter, map, take } from 'rxjs/operators';
+import { EVTStatusService } from 'src/app/services/evt-status.service';
```
ChiaraDipi, line 4: Use relative path

src/app/panels/image-panel/image-panel.component.ts
ChiaraDipi: @szenzaro Can you suggest the better way of doing it?

ChiaraDipi (changes requested):

src/app/panels/image-panel/image-panel.component.ts
szenzaro: Check and review last commit

ChiaraDipi (approved):

szenzaro (approved):

PR #200, opened 2022-11-10
Fix missing space in verses display in dipl transcription

ChiaraDipi:
The words at the end of the line that do not coincide with the end of a line were attached to those of the next line. Now a space has been added.

### DIFF
--- a/src/app/components/lb/lb.component.ts
+++ b/src/app/components/lb/lb.component.ts
@@ -40,6 +40,12 @@ export class LbComponent {
     );
   }
 
+  get displayInline$() {
+    return this.displayBlock$.pipe(
+      map((displayBlock) => !displayBlock),
+    );
+  }
+
   constructor(
     private evtModelService: EVTModelService,
   ) {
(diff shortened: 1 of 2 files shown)


src/app/components/lb/lb.component.ts
```diff
@@ -40,6 +40,12 @@ export class LbComponent {
     );
   }
 
+  get displayInline$() {
+    return this.displayBlock$.pipe(
```
szenzaro, line 44: Make this a properly instead of a getter

szenzaro (changes requested):

src/app/components/lb/lb.component.ts
ChiaraDipi: If I make it an observable it doesn't work any more! :( 
Probably the problem is on the `displayBlock$` property. It is itself a getter and relies on the value of Input properties (`editionLevel` and `textFlow`):
```
get displayBlock$() {
    return this.evtModelService.lines$.pipe(
      map((lines) => lines.length > 0),
      map((hasLines) => {
        if (!this.data.attributes.id && !this.data.attributes.n) {
          return true;
        }
        switch (this.editionLevel) {
          case 'diplomatic':
          case 'interpretative':
            return this.textFlow === 'verses' ? false : hasLines;
          case 'critical':
            return this.textFlow === 'prose';
        }
      }),
    );
  }
```

If I make it a property instead of a getter it doesn't work any more as expected, probably because it doesn't catch changes on Input properties. How can I do?

szenzaro (approved):

PR #220, opened 2023-10-17
Feature/crit apparatus frame

davivcu:
This branch populated the three sections of the Apparatus-Sources-Analogues frame.
Parsing of sources and analogues attributes can be custumized in new keys inside edition_config.json.

The aspect of sources and analogues elements in the text flow can be configured in editorial_conventions_config.json.
Since sources and analogues refer to different markups, said customization will use the dot notation to clarify that we are styling a group of elements '.analogues' and '.sources', and not the single elements quote/cit/div/ref/seg/p/l... that can be recognized as such (example at the end of the post).
 Default values for this editorial configuration joined the similar defaults inside editorial-conventions-service.ts.

The highlight color for click and hover events on the sources and analogues is taken from existing values edition.readingColorDark and edition.readingColorHight key in edition_config.json.
In order to apply them, this branch uses a new small function that can be extended in the future to load generic CSS written by the users.

Both analogue entry and source entry components use the existing evt-content-viewer component in order to display their nested elements both in the text flow and in their frame.
In the case of self-closed elements such as PTR or empty elements, the source or analogue element will be displayed as a note/exponent in the text flow.

_**New features:**_

- Apparatus component in order to list and manage the existing apparatus on the current page.
- Source-entry and source-detail component
- Analogue-entry and analogue-detail component
- Sources list component and analogues (list) component.

Managing Sources and Analogues classes required setting up the parsing and models of various elements:
- BIBL, LISTBIBL, BIBLSTRUCT, and their components
- QUOTE, CIT
- REF, SEG

Added few utilities in dom-utils and xml-utils.

_**Changes to existing features:**_

Elements modified for @source (customizable in config) attribute recognition and relative elements retrieving:
- DIV, P, L, LG both parsing and their component
- NOTE, PTR parsing only

Added correct type referral in MsDescParser for new BiblList, Bibl, and Quote types.

A json you may use to test editorial conventions for analogues and sources:
```
 { 
    "quotations": {
        "markup": {
            "element": ".sources"
        },
        "layouts": {
            "diplomatic": {
                "style": {
                    "font-style": "italic",
                    "background-color":"red"
                }
            },
            "critical": {
                "style": {
                    "font-style": "italic",
                    "background-color":"green"
                }
            },
            "interpretative": {
                "style": {
                    "font-style": "italic",
                    "background-color":"blue"
                }
            }
        }
    },
    "analogues": {
        "markup": {
            "element": ".analogues"
        },
        "layouts": {
            "diplomatic": {
                "style": {
                    "text-decoration": "underline dotted from-font",
                }
            },
            "interpretative": {
        	"pre": '>',
               "style": {
                    "font-style": "italic",
                    "text-decoration": "underline dotted from-font",
                }
            },
            "critical": {
        	 "pre": 'ANG: ',
                "style": {
                    "text-decoration": "underline dotted from-font",
                }
            }
        }
    }
 }
```


src/app/services/evt-status.service.ts
```diff
@@ -179,6 +182,15 @@ export class EVTStatusService {
             params,
         };
     }
+
+    getPageElementsByClassList(classList) {
+        const pageContent = this.currentStatus$.pipe(
+            map(({ page }) => page.parsedContent));
+
+        return pageContent.pipe(map((x) => deepSearch(x, 'class', classList)));
```
ChiaraDipi, line 190: You can have directly a return without the `pageContent` variable. Just pipe `map`s you need to `this.currentStatus$`. 
Moreover, use a more self explaining name instead of `x`

src/app/services/xml-parsers/analogue-parser.ts
```diff
@@ -0,0 +1,130 @@
+import { AppConfig } from 'src/app/app.config';
+import { parse, ParserRegister, xmlParser } from '.';
+import { Analogue, AnalogueClass, BibliographicEntry, BibliographicList, GenericElement, Milestone, XMLElement } from '../../models/evt-models';
+import { getOuterHTML } from '../../utils/dom-utils';
+import { AnchorParser, AttributeParser, GenericElemParser, MilestoneParser } from './basic-parsers';
+import { createParser, getID, parseChildren, Parser } from './parser-models';
+import { chainFirstChildTexts, getExternalElements, normalizeSpaces } from 'src/app/utils/xml-utils';
```
ChiaraDipi, line 7: Use relative imports

src/app/utils/xml-utils.ts
```diff
@@ -36,3 +36,141 @@ export function replaceNotWordChar(textContent: string) {
...
+}
+
+/**
+* Significant text sometimes is split inside two or more text evt-element inside the main one, especially when it contains new line characters.
+* This function returns a string with all the text element chained
+* @param n XMLElement
+* @returns string
+*/
+export function chainFirstChildTexts(elem: XMLElement): string {
+  if (elem === undefined) { return ''};
```
ChiaraDipi, line 56: Add `;` and a space after it

src/app/utils/xml-utils.ts
```diff
@@ -36,3 +36,141 @@ export function replaceNotWordChar(textContent: string) {
...
+  const evtTextElements = {
+    '#text': 'nodeValue',
+    'p': 'textContent',
+  };
+  const evtTextComplexElements = ['choice', 'app', 'l', 'quote', 'p', 'lg'];
```
ChiaraDipi, line 61: EVT logic shouldn't stay in utility generic function. I would probably try to put it in the parser, since it is strictly connected to the xml tei parser and use a property passed to the function to indicate complex elements. @giuliac89 @szenzaro what do you think? How would you do?

src/app/utils/xml-utils.ts
```diff
@@ -36,3 +36,141 @@ export function replaceNotWordChar(textContent: string) {
...
+  return out;
+}
+
+/**
+ * Retrieve and chain textContent of all descendents
+ * @param elem ChildNode
+ * @returns out string
+ */
+export function chainDeepTexts(elem: ChildNode): string {
+  const evtInnerTextElements = ['#text', 'reg', 'corr', 'rdg'];
```
ChiaraDipi, line 75: Same as above

src/app/utils/xml-utils.ts
```diff
@@ -36,3 +36,141 @@ export function replaceNotWordChar(textContent: string) {
...
+  return (markerAttrs.includes(elem.getAttribute('type')));
+}
+
+export function isSource(elem: XMLElement, attrs: string[]): boolean {
+  let validAttrs = false;
+  attrs.forEach((attr) => { if (elem.getAttribute(attr) !== null) { validAttrs = true } });
+
+  return (validAttrs);
+}
+
```
ChiaraDipi, line 114: `isAnalogue` and `isSource` are not generic xml utils, but are specific function strictly connected to analogue and source. Moreover I don't understand how you validate the fact that an element is an analogue or source. Can you explain so that I can try and help you deciding how to edit this part of code

src/app/utils/xml-utils.ts
```diff
@@ -36,3 +36,141 @@ export function replaceNotWordChar(textContent: string) {
...
+ */
+export function getCorrespElement(fromElement: XMLElement, toXMLID: string) {
+  const findID = toXMLID.replace('#','');
+  const sameTypeElements = fromElement.ownerDocument.querySelectorAll<XMLElement>(fromElement.tagName);
+  let found = null;
+  sameTypeElements.forEach((x) => { if (x.getAttribute('xml:id') === findID) { found = x; return }});
+  if (found !== null) {
+    found.appendChild(fromElement.cloneNode());
+  }
+}
```
ChiaraDipi, line 128: This function is not used throughout. If not needed, remove it to avoid dead code

src/app/utils/xml-utils.ts
```diff
@@ -36,3 +36,141 @@ export function replaceNotWordChar(textContent: string) {
...
+ * @param toXMLID string
+ * @returns object { text: string, elements: any }
+ */
+export function getSpanToElements(fromElement: XMLElement, toXMLID: string): { text: string, elements: any } {
+
+  if ((fromElement === null) || (fromElement === undefined) || (toXMLID === null)) {
+    return { text: '', elements: [] };
+  }
+
+  const findID = toXMLID.replace('#','');
```
ChiaraDipi, line 143: The variable name is `toXMLId` so I expect to pass directly the ID, without the `#`. Either you change the var name or move the replace function where you use this util.

src/app/utils/xml-utils.ts
```diff
@@ -36,3 +36,141 @@ export function replaceNotWordChar(textContent: string) {
...
+}
+
+
+/**
+ * Retrieve textContent and elements between the provided element and the one found with the xml:id provided
+ * @param fromElement XMLElement
+ * @param toXMLID string
+ * @returns object { text: string, elements: any }
+ */
+export function getSpanToElements(fromElement: XMLElement, toXMLID: string): { text: string, elements: any } {
```
ChiaraDipi, line 137: Then name of the function does not properly explain what it does. Better something like `getContentBetweenElementAndId`

src/assets/config/edition_config.json
```diff
@@ -105,5 +105,19 @@
...
+	"biblView": {
+		"propToShow": ["author","title","date","editor","publisher","pubPlace","citedRange","biblScope"],
+		"showAttrNames": false,
+		"showEmptyValues": false,
+		"inline": true,
+		"comaSeparated": true,
+		"showMainElemTextContent": true
+	},
+	"analogueMarkers": ["ParallelPassage", "parallelPassage", "Parallel", "parallel"],
+	"sourcesExcludedFromListByParent": [ "desc" ]
```
ChiaraDipi, line 122: Format file

src/assets/i18n/en.json
```diff
@@ -172,5 +172,17 @@
     "moreInfoAboutApp": "More info about the apparatus entry",
     "omit": "omit.",
     "wit": "Wit:",
-    "xml": "XML"
+    "xml": "XML",
+    "criticalApparatus":"Critical Aparatus",
```
ChiaraDipi, line 176: Typo: "Aparatus" instead of "Apparatus"

src/assets/i18n/it.json
```diff
@@ -171,5 +171,17 @@
...
+    "analogues": "Passi Paralleli",
+    "source": "Fonte",
+    "analogue": "Passo Parallelo",
+    "notes":"Note Critiche",
+    "refBibl": "Riferimenti Bibliografici",
+    "quotedSources": "Fonti citate",
+    "quotedText": "Testo",
+    "references": "Riferimenti",
+    "analogueIn": "Passo parallelo in",
+    "corresps": "Corrispondenze"
```
ChiaraDipi, line 186: Format file

src/styles.scss
```diff
@@ -182,3 +182,52 @@ body {
...
+
+evt-analogue-entry evt-paragraph,
+evt-quote-entry evt-paragraph {
+    display: inline-block;
+}
+
+evt-analogue-entry evt-paragraph p,
+evt-quote-entry evt-paragraph p {
+    margin-bottom:0;
+}
```
ChiaraDipi, line 233: Format file

src/app/utils/dom-utils.ts
```diff
@@ -229,3 +229,35 @@ export function getCommonAncestor(node1, node2) {
...
+          counter = counter - 1;
+        } else {
+          console.log('EVT WARN: element is too deep, not searching further in', obj);
+        }
+      }
+    }
+  }
+  search(obj);
+
+  return results;
```
ChiaraDipi, line 261: Can you find a better way? Defining a function within a function is not that good. Maybe you can pass the counter as a param of the search and define search outside

src/app/services/xml-parsers/analogue-parser.ts
```diff
@@ -0,0 +1,130 @@
...
+    'Analogue', 'MsDesc'];
+
+    public parse(analogue: XMLElement): GenericElement|Analogue {
+
+        if (!(this.analogueMarker.includes(analogue.getAttribute('type'))) || (analogue.parentElement.tagName === 'cit')) {
+            // no source/target attribute or inside a cit element: not an analogue to display alone
+            return this.elementParser.parse(analogue)
+        }
+
+        const noteID = ['div','p','l','lg','note'];
```
ChiaraDipi, line 33: Change var name in order to let it be more self explaining.

src/app/services/xml-parsers/analogue-parser.ts
```diff
@@ -0,0 +1,130 @@
...
+    elementParser = createParser(GenericElemParser, parse);
+    attributeParser = createParser(AttributeParser, this.genericParse);
+    biblParser = createParser(BibliographyParser, this.genericParse);
+    milestoneParser = createParser(MilestoneParser, this.genericParse);
+    anchorParser = createParser(AnchorParser, this.genericParse);
+
+    analogueMarker = AppConfig.evtSettings.edition.analogueMarkers;
+    biblAttributeToMatch = AppConfig.evtSettings.edition.externalBibliography.biblAttributeToMatch;
+    elemAttributesToMatch = AppConfig.evtSettings.edition.externalBibliography.elementAttributesToMatch;
+    notNiceInText = ['Note', 'BibliographicList', 'BibliographicEntry', 'BibliographicStructEntry',
```
ChiaraDipi, line 23: Why this name? "notNice"?

src/app/services/xml-parsers/analogues-entries-parser.service.ts
```diff
@@ -0,0 +1,22 @@
+import { Injectable } from '@angular/core';
+import { ParserRegister } from '.';
+import { Analogue, AnalogueClass, XMLElement } from '../../models/evt-models';
+
+@Injectable({
+  providedIn: 'root',
+})
+export class AnalogueEntriesParserService {
+
+  private tagName = `.${AnalogueClass}`;
```
ChiaraDipi, line 10: Tag names don't have the dot. The var name is incorrect. Change it or put the dot where needed.
Moreover those two props are private and used only once. You probably don't really need them, just use the value when needed.

src/app/services/xml-parsers/basic-parsers.ts
```diff
@@ -87,13 +90,28 @@ export class TextParser implements Parser<XMLElement> {
 
 @xmlParser('p', ParagraphParser)
 export class ParagraphParser extends EmptyParser implements Parser<XMLElement> {
+    analogueParser = createParser(AnalogueParser, this.genericParse);
+    quoteParser = createParser(QuoteParser, this.genericParse);
+    sourceAttr = AppConfig.evtSettings.edition.externalBibliography.elementAttributesToMatch;
+    analogueMarkers = AppConfig.evtSettings.edition.analogueMarkers;
+    source = null; analogue = null;
```
ChiaraDipi, line 97: Different props on different lines!

src/app/services/xml-parsers/basic-parsers.ts
```diff
@@ -122,6 +140,11 @@ export class LBParser extends EmptyParser implements Parser<XMLElement> {
 @xmlParser('note', NoteParser)
 export class NoteParser extends EmptyParser implements Parser<XMLElement> {
     attributeParser = createParser(AttributeParser, this.genericParse);
+    analogueParser = createParser(AnalogueParser, this.genericParse);
+    quoteParser = createParser(QuoteParser, this.genericParse);
+    sourceAttr = AppConfig.evtSettings.edition.externalBibliography.elementAttributesToMatch;
+    analogueMarkers = AppConfig.evtSettings.edition.analogueMarkers;
+    source = null; analogue = null;
```
ChiaraDipi, line 147: Different props on different lines

src/app/services/xml-parsers/index.ts
```diff
@@ -27,11 +27,14 @@ export class ParserRegister {
         if (nels.includes(tagName)) {
             return 'evt-named-entities-list-parser';
         }
+        const quote = ['quote', 'cit', 'ref', 'seg', 'div'];
```
ChiaraDipi, line 30: I honestly don't like this approach, in particular for elements like `seg` or `div` which are very generic and not always used for quotes. I would define a parser dedicated for each element and insert inside of it the check: if it's a quote use quote parser. @szenzaro @giuliac89 what do you think?

ChiaraDipi (changes requested):
I'm not finished yet, but you can fix indications I gave you for the moment. 
I didn't analyze deeply the new parsers yet, but as I indicated in a comment I don't really like your approach to use the same quote parser for element like seg, div, ref and other very generic elements. Try a different approach.

Moreover, I give you some general indications (in some file I've also added the comment, but please check every file you changed/added with these indications in mind):
- Definition of @param and @returns is not needed if you don’t give additional information about them
- Format every file before saving: I’ve found several missing spaces throughout the changes
- Alway order alphabetically imports and injection in constructors
- Try to avoid var names like x, y, z, etc
- Try to avoid `any` and always give a definition for properties types (if not possible add comment to disable eslint checks).
- Always remember that EVT is general purpose, thus sometimes some configurations or elements will be undefined. You should always handle undefined stuff with defaults or fallbacks

src/app/services/xml-parsers/index.ts
davivcu: I could either do that or add a disambiguation parser to distinguish from ref, seg, div that aren't sources or analogues and the ones that are. This because the general parser only discriminates on tagName basis which is not sufficient for this purpose. 
That external disambiguation parser we can image (and which was a first solution you can retrieve in first commits) will result in much duplicated code though.

src/app/services/xml-parsers/analogues-entries-parser.service.ts
davivcu: It's a css class because analogues and sources group up different tags

src/app/components/analogues/analogue-detail/analogue-detail.component.html
```diff
@@ -0,0 +1,92 @@
+<div class="source-container" 
+    *ngIf="((analogue.sources.length > 0) || (analogue.extSources.length > 0) 
+    || (analogue.text.length > 0) || (analogue.extLinkedElements.length > 0))">
```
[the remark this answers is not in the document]
davivcu: If the check returns false the element is not shown (no point to show a source without sources, extSources, quoted text or external linked elements.
But i'm ok to move the "visible" logics in a separated property for clarity's sake

src/app/services/editorial-conventions.service.ts
```diff
@@ -125,26 +125,61 @@ export class EditorialConventionsService {
...
-          ...defaultKeys || {},
-          [editionLevel]: {
-            ...defaultKeys ? defaultKeys[editionLevel] : {},
-            ...externalLayouts[editionLevel],
-          },
-        };
-      });
+      layouts = {
+        ...externalLayouts || {},
+      }
```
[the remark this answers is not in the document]
davivcu: Yes that's correct, please see the hotfix we merged to develop just before the new PR

src/app/services/editorial-conventions.service.ts
```diff
@@ -5,7 +5,7 @@ import { EditorialConvention, EditorialConventionLayouts } from '../models/evt-m
 
 // List of handled editorial convention
 export type EditorialConventionDefaults = 'addition' | 'additionAbove' | 'additionBelow' | 'additionInline' | 'additionLeft' | 'additionRight' |
-  'damage' | 'deletion' | 'sicCrux' | 'surplus';
+  'damage' | 'deletion' | 'sicCrux' | 'surplus' | '.sources' | '.analogues' ;
```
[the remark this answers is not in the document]
davivcu: The dot is a mimic feature for Class (like in CSS), I used it for sources and analogues to show to the user that they are not customizing single elements but group of elements  since many markups (included additions, damages, deletions) can be sources or analogues. Please note that dot in the name is properly managed and doesn't cause any issue.

src/app/services/editorial-conventions.service.ts
```diff
@@ -125,26 +125,61 @@ export class EditorialConventionsService {
...
   };
 
   getLayouts(name: string, attributes: AttributesMap, defaultsKey: EditorialConventionDefaults) {
+    const excludedFromAttributeControl = ['.sources', '.analogues'];
     const defaultKeys = this.defaultLayouts[defaultsKey];
     let layouts: Partial<EditorialConventionLayouts> = defaultKeys;
 
     const externalLayouts = this._getExternalConfigs().find((c) => c.element === name &&
-      (!attributes || Object.keys(attributes).concat(
+      (excludedFromAttributeControl.includes(name) || !attributes || Object.keys(attributes).concat(
```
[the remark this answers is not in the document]
davivcu: We don't want to exclude from custom editiorial conventions analogues and sources that have attributes

src/app/utils/xml-utils.ts
davivcu: Analogues are elements with a signature attributes type (es: "parallelPassage") set by the user in edition_config.json 
Sources are elements with one of the attributes (es: "source" or "target") set by the user in edition_config.json

src/app/components/analogues/analogue-detail/analogue-detail.component.html
```diff
@@ -0,0 +1,92 @@
...
+        
+    <div class="card-body app-detail-content source-detail-content" *ngIf="analogue.content.length > 0">
```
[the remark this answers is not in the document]
davivcu: No because the element comes from analogue parser and analogue parser will always at least set the content property with default value [] empty array

src/app/models/evt-models.ts
```diff
@@ -170,6 +170,68 @@ export class ApparatusEntry extends GenericElement {
...
+    series: BibliographicEntry[];
+    originalEncoding: string;
+}
+
+export class Analogue extends GenericElement {
+    id: string;
+    text: string;
+    sources: BibliographicEntry[];
+    extSources: BibliographicEntry[];
+    extLinkedElements: any;
```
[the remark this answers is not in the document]
davivcu: It really can be any element... first i can think of are l, lg, p, ref, seg, cit, div, quote... it depends on what the @target attribute links to

src/app/utils/xml-utils.ts
davivcu: It will be used in new features I will develop soon in another branch but I can remove it for now if you prefer

src/app/services/xml-parsers/analogue-parser.ts
davivcu: Those are the usually nested elements (bibl, notes, ref etc...) that will not be displayed in text flow but only in the analogue-component detail view

src/app/models/evt-models.ts
```diff
@@ -1042,6 +1112,30 @@ export class Term extends GenericElement {
...
+export class Anchor extends GenericElement {
+    id?: string;
+}
+
+export class Span extends GenericElement {
+    id?: string;
+    from: string;
+    to: string;
+    includedText: string;
+    includedElements: any;
```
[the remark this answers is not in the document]
davivcu: Same as analogue the elements included in Span can really be anything is found between the end and the start of the span, it's impossible to know it before

src/app/models/evt-models.ts
```diff
@@ -170,6 +170,68 @@ export class ApparatusEntry extends GenericElement {
...
+    title: string[];
+    date: string[];
+    publisher: string[];
+    pubPlace: string[];
+    citedRange: string[];
+    biblScope: string[];
+    text: string;
+    quotedText: string;
+    insideCit: boolean;
+    originalEncoding: string;
```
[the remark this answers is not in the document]
davivcu: Ok i took as a model ApparatusEntry because I was first requested to show that element in the frame and there originalEncoding type was a string. Should i change it there too? (evt-models.ts line 169)

src/app/utils/xml-utils.ts
davivcu: It's safe-measure to sanitize a provided 'dirty' xml-id, also If I do it outside We will have to do it each time someone calls this function vs this single line

src/assets/config/edition_config.json
davivcu: ok

src/app/services/editorial-conventions.service.ts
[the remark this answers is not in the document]
giuliac89: I agree with Chiara. Even if it doesn't cause any issue it's better to use the same notation.

src/app/services/xml-parsers/analogue-parser.ts
giuliac89: Can you can find a better name?

PR #242, opened 2024-10-02
Bugfix/no recompiling for custom css

davivcu:
- Move custom style css to static sources in order to avoid the need of compilating them
- Set max zoom level value for images in config


RenatoCaenaro (approved):
Fine for me. 
Accepted!

PR #248, opened 2024-10-07
Updates sample document images

davivcu:
Updates sample document images and change a couple of values for config files


RenatoCaenaro (approved):
it's ok

PR #249, opened 2024-10-07
moved image folder under data

RenatoCaenaro:
moved images that was previously added in the wrong path:
"asset\images"
in the correct path:
"assets\data\images"


RobertoRDT (approved):
OK

PR #251, opened 2024-10-07
Change default values for Saba

davivcu:
Change default values for Saba

- Disables image-text view
- Set default text flow to verses


RobertoRDT (approved):
Approved

PR #256, opened 2025-02-09
Feature/accessibility improvement

davivcu:
We can start revision of the accessibility improvements branch.
It impacts various areas so we should merge it before anything else

- Improve UI scaling with base font size in config
- Add required attributes for accessibility standards (such as aria-label etc...)
- Add navigation between pages with keyboard arrows
- Add switching between views with shortcut keys (maybe we can set them in config?)
- Add semantic tags 
- General revision with the help of Silktide accessibility inspection app


RenatoCaenaro (commented):
The lines you commented out in the css and scss files should be deleted later to keep the sources cleaner.

RenatoCaenaro (approved):

PR #263, opened 2025-03-14
Bug fix about diplomatic visualization

RenatoCaenaro:
This PR Contains:
- edition_config syntax fixes to handle the new synoptic view requirements. 
- custom-styles.css fixes to handle correctly lb inside w


RenatoCaenaro:
after testing there is still some issues in visualization, will keep the PR open and integrate it next week

PR #297, opened 2025-08-31
Feature/app correction sequence

davivcu:


RenatoCaenaro (commented):
Why removing .vscode/launch.json? 
it's useful to debug directly from vscode.

davivcu:
> Why removing .vscode/launch.json? it's useful to debug directly from vscode.

I thought I added it myself by mistake so I removed it; it’s a service file specific for an IDE that I usually wouldn’t include in a repository. That said, since we’re about to merge into develop branch and it does no harm, I’m fine with adding it back.

RenatoCaenaro (approved):

PR #332, opened 2026-07-10
Bugfix/explosive word elements

davivcu:
- Managed intentional whitespace between elements
- Reduced performance impact of w elements
- Managed lb inside w


src/app/components/word/word.component.html
RenatoCaenaro, line 1: evt-content-viewer must be used

RenatoCaenaro: evt-content-viewer bring also back apparatus exponents. 
On-hover event on the exponent do not apply anymore understroke style of words, need to investigate this before approval

RenatoCaenaro (approved):

